Issue #5, opened 2014-11-07
io.py: IndexError: list index out of range processing Vanguard 1 example


gorticus:
In a3c8da153f, executing the Vanguard 1 example in `__init__.py`:
```python
>>> from sgp4.earth_gravity import wgs72
>>> from sgp4.io import twoline2rv
>>>
>>> line1 = ('1 00005U 58002B   00179.78495062  '
... '.00000023 00000-0 28098-4 0 4753')
>>> line2 = ('2 00005  34.2682 348.7242 1859667 '
... '331.7664 19.3264 10.82419157413667')
>>>
>>> satellite = twoline2rv(line1, line2, wgs72)
IndexError                                Traceback (most recent call last)
<ipython-input-12-8ec3d1719ba6> in <module>()
----> 1 satellite = twoline2rv(line1, line2, wgs72)

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sgp4/io.pyc in twoline2rv(longstr1, longstr2, whichconst, afspc_mode)
    150        if longstr1[62] == ' ':
    151            longstr1[62] = '0';
--> 152        if longstr1[68] == ' ':
    153            longstr1[68] = '0';
    154

IndexError: list index out of range
```
Works in v1.3 (4266bfc ), but haven't bisected to find culprit.

brandon-rhodes:
I have just released version 1.4, that shows the problem with lines that have been copied with incorrect whitespace. The script at the top of this issue now results in:

```
Traceback (most recent call last):
  File "tmp2.py", line 8, in <module>
    satellite = twoline2rv(line1, line2, wgs72)
  File "/home/brandon/sgp4/sgp4/io.py", line 153, in twoline2rv
    raise ValueError(error_message.format(1, LINE1, line))
ValueError: TLE format error

The Two-Line Element (TLE) format was designed for punch cards and is
therefore very strict about the position of every space and digit in a
TLE line.  Your line does not quite match.  Here is the official format
for line 1 followed by the line you provided:

1 NNNNNC NNNNNAAA NNNNN.NNNNNNNN +.NNNNNNNN +NNNNN-N +NNNNN-N N NNNNN
1 00005U 58002B   00179.78495062  .00000023 00000-0 28098-4 0 4753
```

Enjoy!

PR #6, opened 2014-11-20
Checking for a min length of TLE

nloadholtes:
A TLE is expected to be 69 characters long. This code will take a string, turn it into a list of chars, and make sure it is at least 69 chars.
Longer lines are possible, but only in the test data located in SGP4-VER.TLE

This is for issue #5

### DIFF
--- a/sgp4/io.py
+++ b/sgp4/io.py
@@ -12,7 +12,7 @@ from sgp4.propagation import sgp4init
 
 INT_RE = re.compile(r'[+-]?\d*')
 FLOAT_RE = re.compile(r'[+-]?\d*(\.\d*)?')
-
+TLE_LENGTH = 69
 
 """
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
Inspired by your code, I decided to do something far more thorough: rewrite the parser code so that, if the line does not match the format, it could show the user the problem. Thanks for giving me the shove to do this properly — the fix is in version 1.4, just released!

brandon-rhodes:
(The particular reason your code cannot now merge: I decided to get rid of my silly code that tried to mimic the byte-array operations of the C, and switched to equivalent easy-to-read parsing that throws away the whole scanf() idea! What was I thinking when I wrote that?)

nloadholtes:
No worries! I'm glad I was able to help

PR #13, opened 2016-01-24
Add conda recipe

Juanlu001:
The version string is based now on the master branch because 1.4 does not work with Python 2.6.

### DIFF
--- a/condarecipe/bld.bat
+++ b/condarecipe/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
--- a/condarecipe/build.sh
+++ b/condarecipe/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
--- a/condarecipe/meta.yaml
+++ b/condarecipe/meta.yaml
@@ -1,0 +1,33 @@
+package:
+  name: sgp4
+  version: "1.5.dev0"
+
+source:
+  git_url: https://github.com/brandon-rhodes/python-sgp4.git
+  git_tag: master
+
+build:
+  number: 1
+
+requirements:
+  build:
+    - python
+
+  run:
+    - python
+
+test:
+  imports:
+    - sgp4
+
+  requires:
+    - unittest2 [py26]
+
+about:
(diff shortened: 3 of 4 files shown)


brandon-rhodes:
I'm happy to release a new 1.5 to get the Python 2.6 fixes out into the wild! First, a bit of context on this pull request: is there any way that Conda can pull the version, description, and dependency information from the `setup.py` file? If I merge this pull request, then it looks like I am doomed forever to have to make every change to the project metadata in two places instead of in one place. (Or, just as likely, to forget to make the change in the Conda files and then the two will be out of sync.)

I do see that the Conda file describing how to run the tests is telling Conda something that is not included in the `setup.py` — but is there some way to avoid duplicating all the rest of it? I'm a bit surprised, given how ubiquitous `setup.py` files are, that Conda doesn’t have a switch that says “pull all of this usual stuff from the `setup.py`” — or that Conda doesn't at least have a packager that could take the `.tar.gz` that results from `setup.py sdist` and use its `PKG-INFO` file to get all of the information that it needs (since there it's in declarative, rather than executable, form).

If Conda lakes this ability, I'd almost rather write a tool for them that can turn a Python `.tar.gz` into a Conda package, than have to maintain this whole set of extra text files separately from where I'm already keeping up with this information. (Especially since I maintain several projects and if something does not stop the madness, they soon will all have these files, each repeating exactly what's already in their `setup.py` — so I'm asking because it does not just involve this one package, but a dozen that I maintain!)

Juanlu001:
Hi Brandon, I totally share your concerns - and I fear I cannot be of much help here, since I'm "doomed" for my own projects and keep updating the metadata everywhere.

I created the recipe using `conda skeleton`, which takes a PyPI package and spits an almost working recipe. Then I added the `run_test.sh` file myself and that's it. The problem is, as you point out, that if the dependencies or the version change one has to update them there too. This was discussed in the conda mailing list some months ago:

http://comments.gmane.org/gmane.comp.python.conda.user/1336

One proposed solution was to add an `--update` flag to `conda skeleton`, which would in turn reduce the burden for maintaining two build systems. It hasn't been implemented though.

https://github.com/conda/conda-build/issues/492

I understand if you don't want to merge this, so I'm leaving the PR open anyway with a reference to the original issue in case anyone finds it useful.

Juanlu001:
Hello, I just wanted to point out that I'm shifting my workflow and tend to not maintain conda recipes inside the source repository anymore - instead, we can keep them on their own repository thanks to the conda-forge community process, which integrates the recipes with continuous integration on Linux, OS X and Windows. I made a recipe for python-jplephem last week in case you want to take a look (https://github.com/conda-forge/staged-recipes/pull/214/files) and was pretty happy with the result :)

I'm closing this PR to favor the conda-forge approach, please ping me if you think otherwise. If I ever create a python-sgp4 feedstock there I can add you as a co-maintainer if you wish.

brandon-rhodes:
Thanks for the update, @Juanlu001! Let me know if the library ever fails to build properly for you.

PR #16, opened 2016-07-26
refactor line format checks to be Numba-friendly

ckuethe:
and fix a TODO while I'm here...


brandon-rhodes:
Wow, is it November already? I let other obligations keep me away from GitHub for too long at a time. I need to start making it a part of every week, so that my projects can make progress!

Thanks for this improvement — it moves the code farther from the original C code, but as it passes all the tests that came with the C code, I think it's time that it start becoming more Pythonic. I'll merge it!

PR #31, opened 2018-09-02
Add sgp4.transforms converted from satellite.js

kannkyo:
i want to use useful utility function like satellite.js in python. so i import that.

https://github.com/shashwatak/satellite-js

### DIFF
--- a/sgp4/propagation.py
+++ b/sgp4/propagation.py
@@ -1192,7 +1192,7 @@ def _initl(
              gsto = gsto + twopi;
 
      else:
-        gsto = _gstime(epoch + 2433281.5);
+        gsto = gstime(epoch + 2433281.5);
 
      return (
        no,
@@ -1948,7 +1948,7 @@ def sgp4(satrec, tsince, whichconst=None):
 * --------------------------------------------------------------------------- */
 """
 
-def _gstime(jdut1):
+def gstime(jdut1):
 
      tut1 = (jdut1 - 2451545.0) / 36525.0;
      temp = -6.2e-6* tut1 * tut1 * tut1 + 0.093104 * tut1 * tut1 + \
--- a/sgp4/tests.py
+++ b/sgp4/tests.py
@@ -8,13 +8,15 @@ except:
 import os
 import re
 import sys
+import json
 from doctest import DocTestSuite, ELLIPSIS
 from math import pi, isnan
 
 from sgp4.earth_gravity import wgs72
-from sgp4.ext import invjday, newtonnu, rv2coe
-from sgp4.propagation import sgp4
+from sgp4.ext import invjday, newtonnu, rv2coe, jday
+from sgp4.propagation import sgp4, gstime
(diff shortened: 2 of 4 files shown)


brandon-rhodes:
Thanks for your interest in SGP4 and making satellite data easier for Python users to access!

My guess is that this nicely compact set of routines belongs in a different library than `sgp4` itself. The intention of `sgp4` is to package up for Python programmers exactly the routines available in the official report on the algorithm:

http://www.celestrak.com/publications/AIAA/2006-6753/

It would, I fear, be misleading to start including routines from other libraries across the Internet inside of `sgp4` since those routines were not actually defined in the PDF paper linked to from the web page above.

But there are two improvements we can make to help you with your project to start porting `satellite-js` to Python:

1. First, I can review the `_gstime()` function that you need changed from private to public and see whether we can go ahead and make that change. That would allow your routines to use `sgp4` without accessing any private underscore-prefixed routines.

2. Second, I can update the documentation page for SGP4 to point out other libraries that go beyond simply implementing the standard routines defined in the Spacetrack paper by helping users associate geographic coordinates with satellite locations. I could link to Skyfield, my own library for doing so, and also link the library you create with these new routines you've written up.

I'll update this issue as I make progress on these two items. Thanks again for your interest in SGP4 and satellite tracking from Python!

brandon-rhodes:
All right — I think I've addressed these first two shortcomings that this identified with the documentation and function names in the library. Are there other steps I can take to make SGP4 easy to use by extension routines like the ones that you are porting from JavaScript? (And: once you've released your routines as a Python module of their own, I'll be happy to link to it from the documentation right next to the link for Skyfield.)

brandon-rhodes:
(And I hope, by the way, that you're in no way discouraged by the fact that SGP4 needs, per its mission, to remain a limited implementation of only the algorithms described in the paper — the routines you've ported are still very useful, as I regularly hear from Skyfield users who think that it's too exact and too slow to easily let them track satellites from Python. A sleeker and simpler and less fussy set of routines like the ones you've ported here from JavaScript would be a perfect complement to this `sgp4` library for folks who just want to do satellite tracking, and don't want the expense of Skyfield's exactness when it comes to defining and converting between coordinate systems!)

kannkyo:
Thank you for your kindness reply.

I can understand your design concept of this wonderful library, sgp4. And I agree your opinion.

The reason of this pull request is my hope that access to orbital data more easily from ground segment engineer. In fact, useful, high reliability (and commercial usable) python library about orbital data are very few in the world. I searched libraries about orbital data at github, libraries.io and PyPI. sgp4 was most satisfying my request.

I will try to make my own module. This module will import and use sgp4. if  can make it, it's my first module!

Best regard.

brandon-rhodes:
I'm going to close this for now, but please feel free to add a comment here if your library appears on PyPI; and, even as a closed PR, we can still point people at this code if we learn they need it. Thanks!

PR #32, opened 2019-04-12
Vectorize `tsince` in `propagation.sgp4()`

dsholes:
WARNING: This is a **work in progress**. I just wanted to get some feedback.

This speeds up propagation for large Time arrays in skyfield
`EarthSatellite.at()` method. Ideally, we would no longer need to use a for-loop for
`len(tsince) > 1`. Instead we can use vectorized calculations.

The main change is to use `numpy` math functions instead of the built-in
`math` module, to enable element-wise calculations anywhere that
`tsince` is used.

The other change is all of the `if` checks on things like
`em` (eccentricity) and `inclm` (inclination) need a ndarray.any() call,
e.g. `if em >= 1.0` needs to change to `if (em >= 1.0).any()`
I've only updated a subset of those to enable testing of TLEs of certain
Flock and SkySat satellites, which are in LEO. Obviously, a better way of 
error handling for single time points within the array is necessary as well.

I've ran some speed tests to show the benefit (~50X speed up).

But unfortunately I don't have a way to test accuracy. Upon comparing position vectors output from both non-vectorized and vectorized implementation, the norm of the vectors can differ by up to 0.08% (or up to 5km for some sats tested). This seems like a very small difference, but to be honest, I can't explain why there's any difference at all. The only thing I can think of is the different implementations between the `numpy` module and `math` module leads to very small differences that accumulate throughout the many arithmetic operations in `sgp4`.

I guess what I'm wondering is **have you tried to do this in the past**? And if you did, is there something in the sgp4 code that makes vectorizing the time calcs difficult?

See performance tests below:

```python
%load_ext autoreload
%autoreload 2

# local "vectorized" version of sgp4.propagation
from propagation import sgp4

import sgp4.propagation as prop_actual
from skyfield.api import Topos, load
import numpy as np
```

# Create `skyfield` EarthSats to test

```python
ts = load.timescale()
stations_url = 'http://celestrak.com/NORAD/elements/planet.txt'
satellites = load.tle(stations_url,reload=True,)
sat_keys = [k for k in satellites.keys() if "FLOCK" in str(k)][0:100]

sim_dur_day = 7. # days
sim_dur_min = np.arange(0.,sim_dur_day*24.*60.+1, 1.)
sf_t_arr = ts.utc(year=2019, month=4,day=13,
                  hour=0,minute=sim_dur_min)

min_past_epoch_arr = []
for earthsat in [satellites[k] for k in sat_keys]:
    sat = earthsat.model
    t = sf_t_arr
    min_past_epoch = (t._utc_float() - sat.jdsatepoch) * 1440.0
    min_past_epoch_arr.append(min_past_epoch)
```

    [#################################] 100% planet.txt
    

# Compare speed of vectorized vs. non-vectorized implementation

## For 1 satellite (and 10081 propagation time points)

```python
%%timeit -n 7 -r 3
# Vectorized
position, velocity = sgp4(sat, min_past_epoch_arr[0])
```

    6.19 ms ± 1.63 ms per loop (mean ± std. dev. of 3 runs, 7 loops each)
    

```python
%%timeit -n 7 -r 3
# Non-vectorized
pos2 = []
vel2 = []
for minute in min_past_epoch_arr[0]:
    position2, velocity2 = prop_actual.sgp4(sat, 
                                            minute)
    pos2.append(position2)
    vel2.append(velocity2)
```

    245 ms ± 6.79 ms per loop (mean ± std. dev. of 3 runs, 7 loops each)
    

## For 100 Satellites (and 10081 propagation time points)

```python
%%timeit -n 3 -r 7
# Vectorized
all_pos = []
all_vel = []
for sat_time in min_past_epoch_arr:
    position, velocity = sgp4(sat, sat_time)
    all_pos.append(np.array(position))
    all_pos.append(np.array(velocity))
all_pos = np.array(all_pos)
all_vel = np.array(all_pos)
```

    462 ms ± 46.2 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)
    

```python
%%timeit -n 1 -r 1
# Non-vectorized
all_pos2 = []
all_vel2 = []
for sat_time in min_past_epoch_arr:
    pos2 = []
    vel2 = []
    for minute in sat_time:
        position2, velocity2 = prop_actual.sgp4(sat, 
                                                minute)
        pos2.append(position2)
        vel2.append(velocity2)
all_pos2.append(pos2)
all_vel2.append(vel2)
```

    25.5 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
    

### Just a note:
It's often faster to preallocate numpy arrays if you know the shape of the output data than it is to use `.append`. See below (`436 ms` vs `462 ms` using `append`). Whether the boost is worth the ugliness/inconvenience is up to you.

```python
%%timeit -n 3 -r 7
# Vectorized with preallocated arrays to hold data
all_pos = np.zeros(
    (len(min_past_epoch_arr),
     3,
     len(min_past_epoch_arr[0])
    )
)
all_vel = np.zeros(
    (len(min_past_epoch_arr),
     3,
     len(min_past_epoch_arr[0])
    )
)
for i,sat_time in enumerate(min_past_epoch_arr):
    position, velocity = sgp4(sat, sat_time)
    all_pos[i,:] = np.array(position)
    all_vel[i,:] = np.array(velocity)
```

    436 ms ± 24.2 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)
    

# Accuracy Comparison

```python
# Vectorized
all_pos = np.zeros(
    (len(min_past_epoch_arr),
     len(min_past_epoch_arr[0]),
     3
     )
)
all_vel = np.zeros(
    (len(min_past_epoch_arr),
     len(min_past_epoch_arr[0]),
     3
     )
)
for i,sat_time in enumerate(min_past_epoch_arr):
    position, velocity = sgp4(sat, sat_time)
    all_pos[i,:] = np.array(position).T
    all_vel[i,:] = np.array(velocity).T
```

```python
# Non-vectorized
all_pos2 = []
all_vel2 = []
for sat_time in min_past_epoch_arr:
    pos2 = []
    vel2 = []
    for minute in sat_time:
        position2, velocity2 = prop_actual.sgp4(sat, 
                                                minute)
        pos2.append(position2)
        vel2.append(velocity2)
    all_pos2.append(np.array(pos2))
    all_vel2.append(np.array(vel2))
```

# Need a better way to test accuracy (see below)

```python
from vectorlib import norm
```

```python
new_sgp4 = all_pos[0]
old_sgp4 = np.array(all_pos2)[0]
```

# Max. Percent difference in magnitude of vectors

```python
(100.*(norm(old_sgp4) - norm(new_sgp4))/norm(old_sgp4)).max()
```

    0.08387164664252654

# Max. difference (in km) between magnitude of vectors

```python
np.abs(norm(old_sgp4) - norm(new_sgp4)).max()
```

    5.76860458667943

### DIFF
--- a/sgp4/propagation.py
+++ b/sgp4/propagation.py
@@ -30,7 +30,9 @@ except ImportError:
                     return jit(jit_this, **jit_options)
                return partial_fake_jit
 
-from math import atan2, cos, fabs, pi, sin, sqrt
+#from math import atan2, cos, fabs, fmod, pi, sin, sqrt
+from numpy import cos, fabs, fmod, pi, sin, sqrt, where
+from numpy import arctan2 as atan2
 
 deg2rad = pi / 180.0;
 _nan = float('NaN')
@@ -160,7 +162,7 @@ twopi = 2.0 * pi
   ----------------------------------------------------------------------------*/
 """
 
-def _dpper(satrec, inclo, init, ep, inclp, nodep, argpp, mp, afspc_mode):
+def _dpper(satrec, inclo, init, ep, inclp, nodep, argpp, mp, opsmode):
 
      # Copy satellite attributes into local variables for convenience
      # and symmetry in writing formulae.
@@ -279,17 +281,17 @@ def _dpper(satrec, inclo, init, ep, inclp, nodep, argpp, mp, afspc_mode):
            dbet   = -ph * sinop + pinc * cosip * cosop;
            alfdp  = alfdp + dalf;
            betdp  = betdp + dbet;
-           nodep  = nodep % twopi if nodep >= 0.0 else -(-nodep % twopi)
+           nodep  = fmod(nodep, twopi);
            #   sgp4fix for afspc written intrinsic functions
            #  nodep used without a trigonometric function ahead
-           if nodep < 0.0 and afspc_mode:
+           if nodep < 0.0 and opsmode == 'a':
                nodep = nodep + twopi;
            xls = mp + argpp + pl + pgh + (cosip - pinc * sinip) * nodep
            xnoh   = nodep;
            nodep  = atan2(alfdp, betdp);
            #   sgp4fix for afspc written intrinsic functions
            #  nodep used without a trigonometric function ahead
-           if nodep < 0.0 and afspc_mode:
+           if nodep < 0.0 and opsmode == 'a':
                nodep = nodep + twopi;
            if fabs(xnoh - nodep) > pi:
              if nodep < xnoh:
@@ -941,8 +943,6 @@ def _dsinit(
   ----------------------------------------------------------------------------*/
 """
 
-@jit(cache=True)
-#@jit
 def _dspace(
        irez,
        d2201,  d2211,  d3210,   d3222,  d4410,
@@ -1138,7 +1138,7 @@ def _initl(
        satn,      whichconst,
        ecco,   epoch,  inclo,   no,
        method,
-       afspc_mode,
+       opsmode,
        ):
 
      # sgp4fix use old way of finding gst
@@ -1175,7 +1175,7 @@ def _initl(
      method = 'n';
 
      #  sgp4fix modern approach to finding sidereal time
-     if afspc_mode:
+     if opsmode == 'a':
 
          #  sgp4fix use old way of finding gst
          #  count integer number of days from 0 jan 1970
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thanks for thinking of ways to try to make the library faster! Could the problem be that the `if any(...)` approach means that an adjustment that normally would only be made for one of the satellites matching the `if` condition will now be made to all of them, throwing the other ones off?

dsholes:
Hi Brandon,

I found a bug. I'm using `np.where()` for the if any() cases, but for the `em` line, I switched the order of the arguments:

`em = np.where(em < 1.0e-6, em, 1.0e-6);` should be
`em = np.where(em < 1.0e-6, 1.0e-6, em);`

That actually fixed everything. Now the results agree exactly for my tests. 

I realized however that I edited an old version of `sgp4.propagation`, so I've re-forked, deleted my pull request, and I'll make a new one after I've finished the edits.

I imagine more work will still need to be done for when `sat.method=='d'`. I've only touched `propagation.sgp4()`

PR #33, opened 2019-04-12
Vectorize `tsince` in `propagation.sgp4()`

dsholes:
This speeds up propagation for large Time arrays in skyfield
`EarthSatellite.at()` method. No longer need to use for-loop for
`len(tsince) > 1`. Instead can use vectorized calculation.

WARNING: This is a work in progress.

The main change is to use `numpy` math functions instead of the built-in
`math` module, to enable element-wise calculations anywhere that
`tsince` is used.

The other change is all of the `if` checks on things like
`em` (eccentricity) and `inclm` (inclination) need a ndarray.any() call,
e.g. `if em >= 1.0` needs to change to `if (em >= 1.0).any()`, and to
use `np.where` to replace certain `if-else` statements.

More work must be done for other methods within `sgp4.propagation`.
I've only touched `propagation.sgp4`. I imagine things will break if
sat.method = 'd' (functions related to deep space). A better way of
error handling for single time points within the array is necessary as
well.

I've only updated a subset of those to enable testing of TLEs of certain
Flock and SkySat satellites, which are in LEO.

I've ran some speed tests to show the benefit (~50X speed up).

See below for performance tests:

<details>

```python
%load_ext autoreload
%autoreload 2

# local "vectorized" version of sgp4.propagation
from propagation import sgp4

import sgp4.propagation as prop_actual
from skyfield.api import Topos, load
import numpy as np
```

# Create `skyfield` EarthSats to test

```python
ts = load.timescale()
stations_url = 'http://celestrak.com/NORAD/elements/planet.txt'
satellites = load.tle(stations_url,reload=True,)
sat_keys = [k for k in satellites.keys() if "FLOCK" in str(k)][0:100]

sim_dur_day = 7. # days
sim_dur_min = np.arange(0.,sim_dur_day*24.*60.+1, 1.)
sf_t_arr = ts.utc(year=2019, month=4,day=13,
                  hour=0,minute=sim_dur_min)

min_past_epoch_arr = []
sat_models = []
for earthsat in [satellites[k] for k in sat_keys]:
    sat = earthsat.model
    t = sf_t_arr
    min_past_epoch = (t._utc_float() - sat.jdsatepoch) * 1440.0
    sat_models.append(sat)
    min_past_epoch_arr.append(min_past_epoch)
```

    [#################################] 100% planet.txt
    

# Compare speed of vectorized vs. non-vectorized implementation

## For 1 satellite (and 10081 propagation time points)

```python
%%timeit -n 1 -r 1
# Vectorized
position, velocity = sgp4(sat_models[0], min_past_epoch_arr[0])
```

    9.3 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
    

```python
%%timeit -n 1 -r 1
# Non-vectorized
pos2 = []
vel2 = []
for minute in min_past_epoch_arr[0]:
    position2, velocity2 = prop_actual.sgp4(sat_models[0], 
                                            minute)
    pos2.append(position2)
    vel2.append(velocity2)
```

    246 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
    

## For 100 Satellites (and 10081 propagation time points)

```python
%%timeit -n 1 -r 1
# Vectorized
all_pos = []
all_vel = []
for i,sat_time in enumerate(min_past_epoch_arr):
    position, velocity = sgp4(sat_models[i], sat_time)
    all_pos.append(np.array(position))
    all_pos.append(np.array(velocity))
all_pos = np.array(all_pos)
all_vel = np.array(all_pos)
```

    590 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
    

```python
%%timeit -n 1 -r 1
# Non-vectorized
all_pos2 = []
all_vel2 = []
for i,sat_time in enumerate(min_past_epoch_arr):
    pos2 = []
    vel2 = []
    for minute in sat_time:
        position2, velocity2 = prop_actual.sgp4(sat_models[i], 
                                                minute)
        pos2.append(position2)
        vel2.append(velocity2)
all_pos2.append(pos2)
all_vel2.append(vel2)
```

    26.3 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
    

### Just a note:
It's often faster to preallocate numpy arrays if you know the shape of the output data than it is to use `.append`. See below vs. above using`append`. Whether the boost is worth the ugliness/inconvenience is up to you.

```python
%%timeit -n 3 -r 7
# Vectorized with preallocated arrays to hold data
all_pos = np.zeros(
    (len(min_past_epoch_arr),
     3,
     len(min_past_epoch_arr[0])
    )
)
all_vel = np.zeros(
    (len(min_past_epoch_arr),
     3,
     len(min_past_epoch_arr[0])
    )
)
for i,sat_time in enumerate(min_past_epoch_arr):
    position, velocity = sgp4(sat_models[i], sat_time)
    all_pos[i,:] = np.array(position)
    all_vel[i,:] = np.array(velocity)
```

    569 ms ± 8.87 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)
    

# Accuracy Comparison

```python
# Vectorized
all_pos = np.zeros(
    (len(min_past_epoch_arr),
     len(min_past_epoch_arr[0]),
     3
     )
)
all_vel = np.zeros(
    (len(min_past_epoch_arr),
     len(min_past_epoch_arr[0]),
     3
     )
)
for i,sat_time in enumerate(min_past_epoch_arr):
    position, velocity = sgp4(sat_models[i], sat_time)
    all_pos[i,:] = np.array(position).T
    all_vel[i,:] = np.array(velocity).T
```

```python
# Non-vectorized
all_pos2 = []
all_vel2 = []
for i,sat_time in enumerate(min_past_epoch_arr):
    pos2 = []
    vel2 = []
    for minute in sat_time:
        position2, velocity2 = prop_actual.sgp4(sat_models[i], 
                                                minute)
        pos2.append(position2)
        vel2.append(velocity2)
    all_pos2.append(np.array(pos2))
    all_vel2.append(np.array(vel2))
```

# Need a better way to test accuracy (see below)

```python
from vectorlib import norm
```

```python
new_sgp4 = all_pos[0]
old_sgp4 = np.array(all_pos2)[0]
```

# Max. Percent difference in magnitude of vectors

```python
(100.*(norm(old_sgp4) - norm(new_sgp4))/norm(old_sgp4)).max()
```

    9.122555514414154e-14

# Max. difference (in km) between magnitude of vectors

```python
np.abs(norm(old_sgp4) - norm(new_sgp4)).max()
```

    6.366462912410498e-12

</details>


dsholes:
Looks like it's just failing for Python 2.7...

dsholes:
I just noticed an issue. Using `@jit` when `tsince` is a numpy array is very slow the first time `sgp4` is called. I think it's related to this [stackoverflow post](https://stackoverflow.com/questions/50658884/why-this-numba-code-is-6x-slower-than-numpy-code).

In short, it seems like creating a lot of numpy arrays within a numba function is not a good idea. However, numpy outperforms a numba function by an order of magnitude for the case of sgp4 looping over a large `tsince` array.

Thus, for a vectorized version using numpy arrays, it is best to avoid using `@jit`. I'll update the pull request.

brandon-rhodes:
Alas, that I still haven't spent time with this pull request from April! But my schedule now lets me jump in.

My big question about turning this into vectors is the same question I see that you've asked in a comment: what happens with an `if` statement that adjusts parameters, but should only do so for satellites in particular circumstances? I think the PR would need to become more sophisticated in those circumstances, doing tricks like I learned to do in Skyfield where a true/false bitmask applies an adjustment to some satellites but not others.

First: what this PR really needs is a comprehensive test. The test needs to get all the satellites in the test suite data set, from the satellite file that came with SGP4; it needs to turn them into a single big vector of all those satellites; and it needs to send them all at once through SGP4. If they all come out with the positions expected, then we'll have confidence that the vector routines are behaving.

Could you add such a test to the PR and report back about how many satellites currently give good positions when all are submitted at the same time as a vector? Thanks!

dsholes:
The "satellite file" you're referring to is ["SGP4-VER.TLE"](https://github.com/brandon-rhodes/python-sgp4/blob/master/sgp4/SGP4-VER.TLE), correct?

I'll work on the test you're describing. 

The one difference is I won't send all satellites through as a big vector. Based on the way `sgp4` and the `skyfield` earth sat `.at()` methods currently function (treating satellites as objects), I think it makes more sense to continue running one satellite at a time (for-loop method).  The main difference between the normal code and this PR is that the **time array** can be input as a vector, not satellite objects.

But I'll run all TLEs (satellites) in ["SGP4-VER.TLE"](https://github.com/brandon-rhodes/python-sgp4/blob/master/sgp4/SGP4-VER.TLE), and use the same large time vector for all. I'll then check agreement with the normal sgp4 positions, which uses a for loop to move through singular times.

Should the time array include times before the epoch of each satellite (negative tsince)?

Also regarding this:
> where a true/false bitmask applies an adjustment to some satellites but not others.

I've been using `np.where(bool_mask, value_if_true, value_if_false)` whenever I encounter an if-else statement that becomes vectorized with `t`. I imagine that's what you're referring to.

brandon-rhodes:
Did you ever get results from testing your vectorized algorithm against SGP4-VER.TLE? In the meantime, I should note that conversations have advanced about whether this package, or another one, should in addition to this Python SGP4 implementation, wrap the official C implementation to provide extra speed for users needing lots of satellites. So, if it turns out your improvements here struggle to correctly propagate lots of satellites, we maybe could use the C version as an alternative way of producing some speedup for users who need it.

dsholes:
Thanks for the reminder, I got back to this yesterday. The short answer is, the vectorization is working for `satrec.method == 'n'` (near earth), but it would take a big effort and massive changes to the code to get it working for `satrec.method == 'd'` (deep space).

For now, if it's `'d'` I'd stick to the for-loop, and for `'n'` I'd use the vectorized code. I use SGP4 exclusively for LEO sats so this wouldn't be a deal breaker for me, but from a maintainability standpoint, I think it would be much easier / cleaner to go the Cython route. I'm very curious how the speed will compare for near earth cases.

I'm not very familiar with Cython, so sorry if this doesn't make any sense, but would the for-loop for a large time array be "wrapped" so that it's implemented in C instead of in Python? I guess what I'm asking is, if I pass a large time array to sgp4 or skyfield, will the for-loop still be run in Python, and only the calculations take place in the C code, or will everything take place "inside" the C code?

Also, I'm struggling to integrate everything into the current `sgp4.tests` (it's difficult since the tests were written for single "time" calls not, arrays of time). In case you're interested in the vectorized code and my attempt at integrating it into your existing tests, I can push my changes thus far, but fair warning, it's a bit of a mess.

brandon-rhodes:
If the vectorized version only works for one class of satellites, then my guess, alas, is that it would be too confusing for users to merge it into the main package — after all, satellites do not come labelled in the TLE file with whether they are deep space or not; it's the SGP4 import routine that figures that out based on the orbital parameters.

Over the past couple of days I have tried making a minimal wrapper around the raw C++ code from the maintainer for SGP4 itself, and its performance is promising. A couple of years ago I would not have had any confidence in getting a C-language module pushed to users, but today there are much better mechanisms for building Windows and Mac (and Linux) packages using CI services and then getting them all uploaded to PyPI.

Anyway, thanks VERY much for the work you did here to see if the algorithm, without a huge rewrite, would be susceptible to processing vectors of input dates. It's unfortunate that the answer looks to be "no", but that gives us very useful signal about where to take the project next!

Should we close this pull request, or are there some last details you'd like to discuss first?

dsholes:
In my latest commit I added [new tests](https://github.com/brandon-rhodes/python-sgp4/blob/809ca0feffcdd4e516e530ce5cbe6ad13d455f07/test_ds_main.ipynb), in case you're ever interested in integrating these kinds of changes in the future.

Just to clarify, I hide the logic for `'d'` vs `'n'` and single `tsince` vs array of `tsince` within `propagation.py`, so it would never fall on the user to make that decision. 

However, I still agree that I think it makes a lot more sense to wrap the C++ code. You may want to take a look at the [tests](https://github.com/brandon-rhodes/python-sgp4/blob/809ca0feffcdd4e516e530ce5cbe6ad13d455f07/test_ds_main.ipynb) (plots at the bottom of the notebook) to see the performance difference for large time arrays. I would be interested in running a similar test in the future to see how the Cython performance gain compares to the vectorized (numpy) code.

So yes, feel free to close this pull request, and once again thank you so much for all of your work on this library and skyfield. They're both amazing.

brandon-rhodes:
> So yes, feel free to close this pull request, and once again thank you so much for all of your work on this library and skyfield. They're both amazing.

Thanks, it's very encouraging to hear when the libraries have helped folks with their projects!

Issue #19, opened 2017-08-22
When is it time to reformat the code? 


ckuethe:
The reference implementation of sgp4 hasn't changed significantly in a decade. It should be pretty straightforward (if tedious) to reimplement future changes in python, especially if they provide test cases.

To that end, I'm starting to rename variables but we should probably have a discussion about when to pep8 the code. 

https://github.com/ckuethe/python-sgp4/commits/more_pythonic

brandon-rhodes:
My tendency would be to start morphing the code only as we need to touch it for other reasons. One possible reason would be to eliminate the awkward way that we have to call routines with dozens of values all in a tuple — fixing that cleanly would involve touching lots of code that could be cleaned up.

But I'm not sure we should rename any variables, because it would make it very difficult to compare the code with other reference implementations, wouldn't it?

ckuethe:
I wouldn't say "very difficult", just a little bit tedious. But less tedious than wondering what `method = 'd'` does. Diff the upstream C sources and translate to python. If you look at what changed when the "improved" mode arrived (compared to "afspc" mode) the change is pretty small.

In my head, having a pure-python implementation of implies being programmer friendly and pythonic rather than just prepending an appropriate number of spaces to every line. Otherwise, why not just use a python binding to libastro?

brandon-rhodes:
The Python binding to libastro requires someone compiling the library under Windows, which proved a huge problem with PyEphem — so the main point of this SGP4 module was not to try to rewrite all the already-working code to follow different idioms, but to make the minimum number of changes necessary so that it could run under Python. I was not really planning on us spending time in the code, as I suspect it should stay identical in function to the reference SGP4 implementation?

I had supposed that these tweaks we are making to fix Numba compatibility, along with whatever other tweaks we might need to get high performance, would be the last times we would touch the library — once it's a performant port of the original code, I think we can call it done and move back to building on it further in Skyfield? (Which *IS* where I wanted to make everything beautiful and Pythonic!)

ckuethe:
So a black box, optionally accelerated by another black box if you happen to have numba available? And people should don their eclipse goggles and welding masks before attempting to gaze upon the SGP4 code?

brandon-rhodes:
[...]

PR #35, opened 2019-07-29
Bring current with 7 dec 15 vallado CPP version

interplanetarychris:
io.py
  1. Rename afspc_mode to opsmode,
      take a/i as input instead of True/False
  2. Clarify which mean anomaly (no) is used,
      satrec.no variable renamed to satrec.no_kozai,
      internal updates to support name change
  3. Make more internal variables available available to satrec

propagation.py
  1. Rename afspc_mode to opsmode,
      take a/i as input instead of True/False
  2. Initialize whichconst variables internally once,
      make values accessible in satrec
  3. delete unused variables in initl

test.py
  1. Replace deprecated assertRaisesRegexp with assertRaisesRegex

Notes:
  - Confirmed SGP4-VER.TLE is still current
  - Confirmed unit tests pass with
     python -m unittest discover sgp4

### DIFF
--- a/sgp4/io.py
+++ b/sgp4/io.py
@@ -81,7 +81,7 @@ with an N where each digit should go, followed by the line you provided:
 *    typerun     - type of run                    verification 'v', catalog 'c', 
 *                                                 manual 'm'
 *    typeinput   - type of manual input           mfe 'm', epoch 'e', dayofyr 'd'
-*    afspc_mode  - True for afspc calculations, False for 'improved' mode
+*    opsmode     - mode of operation afspc or improved 'a', 'i'
 *    whichconst  - which set of constants to use  72, 84
 *
 *  outputs       :
@@ -99,7 +99,7 @@ with an N where each digit should go, followed by the line you provided:
   --------------------------------------------------------------------------- */
 """
 
-def twoline2rv(longstr1, longstr2, whichconst, afspc_mode=False):
+def twoline2rv(longstr1, longstr2, whichconst, opsmode='i'):
     """Return a Satellite imported from two lines of TLE data.
 
     Provide the two TLE lines as strings `longstr1` and `longstr2`,
@@ -112,7 +112,7 @@ def twoline2rv(longstr1, longstr2, whichconst, afspc_mode=False):
 
     Normally, computations are made using various recent improvements
     to the algorithm.  If you want to turn some of these off and go
-    back into "afspc" mode, then set `afspc_mode` to `True`.
+    back into "opsmode" mode, then set `opsmode` to `a`.
 
     """
 
@@ -139,8 +139,9 @@ def twoline2rv(longstr1, longstr2, whichconst, afspc_mode=False):
         line[63] == ' '):
 
         _saved_satnum = satrec.satnum = int(line[2:7])
-        # classification = line[7] or 'U'
-        # intldesg = line[9:17]
+        satrec.line1 = line
+        satrec.classification = line[7] or 'U'
+        satrec.intldesg = line[9:17]
         two_digit_year = int(line[18:20])
         satrec.epochdays = float(line[20:32])
         satrec.ndot = float(line[33:43])
@@ -148,8 +149,8 @@ def twoline2rv(longstr1, longstr2, whichconst, afspc_mode=False):
         nexp = int(line[50:52])
         satrec.bstar = float(line[53] + '.' + line[54:59])
         ibexp = int(line[59:61])
-        # numb = int(line[62])
-        # elnum = int(line[64:68])
+        satrec.ephtype = line[62]
+        satrec.elnum = int(line[64:68])
     else:
         raise ValueError(error_message.format(1, LINE1, line))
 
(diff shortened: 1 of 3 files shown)


interplanetarychris:
I noticed I couldn't regenerate a TLE from the satrec.model variables (particularly, mean motion (no) was off.  After going down a rabbit hole to determine the problem, I noticed David Vallado had updated the SGP4.CPP to address the issue, as well as other updates.

I'm not sure if the sgp4.ext julian day routines needed updating, given the python implementation of fractions, so I've left those as-is.

brandon-rhodes:
Thanks for noticing a discrepancy between this routine and the most up-to-date one! I'm traveling at the moment for a Python conference, but will plan to tackle this later this month when I'm done.

brandon-rhodes:
Could you link to the source code that Vallado released? I did a quick web search but can't find the source of these changes. Thanks!

ckuethe:
https://www.celestrak.com/software/vallado-sw.php

ckuethe:
Also, as noted in https://github.com/brandon-rhodes/python-sgp4/issues/19#issuecomment-324434570 you can use the wayback machine to find previous versions of `cpp.zip`

interplanetarychris:
Updated test.py to handle deprecated assertRaisesRegexp for Python 2.6/2.7

brandon-rhodes:
> Rename afspc_mode to opsmode, take a/i as input instead of True/False

I note that this rolls back a5174cd321cfa012aec8c09b6eb4ec4e29cd70d1 which was part of trying to make the code more susceptible to Numba acceleration. The description of that commit correctly notes that it “does introduce an API change” and probably because of that warning, I have never yet released a version of this package that includes the change.

If Numba is not going to prove to be our perfect answer for good performance, then I would rather stick as close to Vallado as possible, for folks who are already familiar with his version of each of these routines. So I'm inclined to accept a regression here back to the Vallado-style version of this option, and from now on to focus on (gulp!) a compiled version of the package to supply performance where people need it (for which we have other PR’s currently ongoing).

@ckuethe — I'm probably going to merge this soon, which will, alas, roll back some of the attempts you made to make things Numba-friendly. Please comment here if you think that’s a wrong direction for me to take; I’m open to considering other options. But at the moment my inclination would be to merge this by sometime next week.

ckuethe:
I don't really care about the API to much, I just want it to go fast.

interplanetarychris:
@brandon-rhodes I've been using this fork, along with the cython accelerations from PR https://github.com/brandon-rhodes/python-sgp4/pull/38 for the back-end of https://trusat.org/ I also have a working cthonized/tsince-vectorized fork at https://github.com/interplanetarychris/python-sgp4/tree/7-dec-15-vallado-tsince-vectorize but sadly it does not accelerate performance much for my application of ~10s of points for a single satellite.

I will be putting out a https://gitcoin.co/ or similar bounty in hopes of incentivizing someone in the world to get the functions working for a 10-100x speed bump per my and @ckuethe's need for speed.  Depending on the results of that, it may make sense to maintain a separate "performance" fork of python-sgp4, or depending on the changes, roll them into the main repo.  We shall see...  I will post info about that bounty task here when I have it online - it will include a fork of the repo along with test/performance metric code to help a positive outcome.

brandon-rhodes:
> I also have a working cthonized/tsince-vectorized fork at https://github.com/interplanetarychris/python-sgp4/tree/7-dec-15-vallado-tsince-vectorize but sadly it does not accelerate performance much for my application of ~10s of points for a single satellite.

How does its behavior compare with PyEphem (if you have time to check, of course)? Several folks have been happy with PyEphem through the years, but I can't imagine why it would be any faster than the same routine in Cython.

sgp4/io.py
```diff
@@ -139,17 +139,18 @@ def twoline2rv(longstr1, longstr2, whichconst, afspc_mode=False):
         line[63] == ' '):
 
         _saved_satnum = satrec.satnum = int(line[2:7])
-        # classification = line[7] or 'U'
-        # intldesg = line[9:17]
+        satrec.line1 = line
```
brandon-rhodes: Was this debugging code, or an idea for a new feature? Vallado's routines don't save the lines, and here in Python there may be a danger: will users do things like take a satellite, adjust a parameter, and expect `.line1` to now reflect the new parameter? It might be safer to not cache this data since the source library does not.

brandon-rhodes:
It happens that I have some time today to spend on the library, so I'll make the suggested changes myself post-merge. Thanks again for noticing that the algorithm was out of date!

brandon-rhodes:
Oh, and, in case other followers of this thread have not seen it, there now seems to be a fast compiled vectorized SGP4 for Python:

https://pypi.org/project/cysgp4/

I am going to experiment with using its results in Skyfield, in which case I'll be happy to recommend its use in place of this pure-Python `sgp4` library for folks needing to propagate lots of satellites.

brandon-rhodes:
And I've opened an issue asking some questions about that library, if anyone wants to follow along:

https://github.com/bwinkel/cysgp4/issues/8

PR #40, opened 2019-10-01
Subclass Satellite and use twoline2rv as its constructor

jesseli2002:
I was figuring that this would be helpful if you needed to subclass Satellite and change the __init__, e.g. adding some other property. I've just added a Satellite2 class that subclasses Satellite and pulls all of its attributes from a dummy object generated by twoline2rv. (I imagine Satellite2 isn't a very good name but I couldn't think of anything too descriptive). I figure this keeps it backwards compatible.

I'm going to apologize - I'm a hobby programmer at best, so I'm not really familiar with some of the technologies used (e.g. CI, or how Python manages the source of a package). Nor do I know a lot about astronomy [I'm looking through these packages for a GPS-tracking project]. I've done my best to try to follow the rules, but I apologize if I've broken something, or if this addition isn't very useful.

### DIFF
--- a/sgp4/sat2.py
+++ b/sgp4/sat2.py
@@ -1,0 +1,14 @@
+from sgp4.io import twoline2rv
+from sgp4.model import Satellite
+from sgp4.earth_gravity import wgs72
+
+class Satellite2(Satellite):
+    """
+    Subclass of Satellite that uses the twoline2rv function as a constructor.
+
+    This is useful if you're subclassing Satellite and need to modify the constructor (e.g. attach another attribute)
+    """
+    def __init__(self, longstr1, longstr2, whichconst, afspc_mode=False):
+        _Satellite = twoline2rv(longstr1, longstr2, whichconst, afspc_mode)
+        for attr in vars(_Satellite):
+            setattr(self, attr, vars(_Satellite)[attr])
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
> I'm a hobby programmer at best, so I'm not really familiar with some of the technologies used

Then I'm impressed that your pull request includes a test — most beginners with Python and CI don’t even realize tests are there, or know how to add more!

The pull request, though, raises three questions as I read it.

1. In general I only make software more complicated once a demand arrives. So instead of adding a class like this right now, because someone someday might need to add a property, I think it’s very important to wait for someone with that actual need to come along. First, because otherwise software winds up with a hundred conveniences attached that never get used. And, second, because often the work we do to prepare for a hypothetical future case often doesn’t wind up being the actual solution needed — you can’t know until the actual need comes up. (But if you do have a present-day problem you're trying to solve, I'm happy to hear about it and discuss a solution!)

2. If someone wanted to add a property to a Satellite, in Python they could simply call `sat = twoline2rv()` and then say `sat.weight = 1032` or something like that. A subclass would not be necessary.

3. Builders like `twoline2rv()` are a nice design pattern in Python, and it's often nice to keep classes innocent of the details of how they get built. It's harder to test classes that have lots of setup logic in their `__init__()` methods.

If we do run into this case in the future, a simpler approach might be to add a keyword argument to `twoline2rv()` that specifies which class should be instantiated. It could default to `Satellite`. Or we could have the argument provide a pre-built satellite object for `twoline2rv()` to use, but if it's `None` then `twoline2rv()` would build the `Satellite()` it currently builds.

Hmm. I wonder which approach of those two would be best?

Issue #20, opened 2017-09-30
Return state for decayed satellites?


anyeli:
In your implementation of sgp4, in propagation.py, you have the following "fix":

     #  sgp4fix for decaying satellites
     if mrt < 1.0:

         satrec.error_message = ('mrt {0:f} is less than 1.0 indicating'
                                 ' the satellite has decayed'.format(mrt))
         satrec.error = 6;
         return false, false;

Returning (false,false) discards the calculated values, a behavior I find to be rather unhelpful - although the state is invalid, it is still useful, e.g. for solving for impact time and location via Newton's method.  Could you return the calculated values instead?  You are already indicating the error condition via error and error_message.

brandon-rhodes:
That's an interesting question. In the C language SGP4 code that this Python implementation is based on, the result is returned as:
https://celestrak.com/publications/AIAA/2006-6753/AIAA-2006-6753-Rev2.pdf
```
 if (mrt < 1.0)
 {
// printf("# decay condition %11.6f \n",mrt);
 satrec.error = 6;
 return false;
 } 
```
The problem with returning anything other than `false` is that normal users would think that the satellite were still flying, and would expect it to be at exactly the returned coordinates, because in those cases where the algorithm does not return `false`, users think that they can use the position as a real position for the satellite.

Can you adjust your algorithm so that it searches for the moment when the satellite transitions from a valid SGP4 position to the moment where this routine returns `false`, and compute your re-entry from there? If you wait until a moment when this returns `false`, then I think you are already past the point where SGP4 can return good positions, and you will be running Newton's method on an invalid starting point.

anyeli:
The Celestrak implementation of SGP4 returns false, but it still returns the values through array arguments.  This seems intentional to me; otherwise they could have done this check immediately after calculating mrt and avoided calculating the Cartesian state in the first place.

Newton's method works fine inside the Earth.  The function doesn't become discontinuous at the surface of the Earth.  Of course I can write hacks and brute force methods to work around a programming interface that loses information, but I don't because it's silly - I can just delete one line of code in my copy of an open source library, and I don't even need to recompile anything.  I would imagine anyone else doing serious work would be tempted to do the same.

But I want to be cooperative.  Perhaps you could add an optional "advanced" switch, either stored in the Satellite object or passed as an optional argument to propagate and sgp4 for users who don't want to lose the value inside the Earth and actually check the error code and/or perform their own checks?

brandon-rhodes:
[...]

PR #41, opened 2019-12-06
Fix mrt check to return NAN

dsholes:
When writing tests for the vectorized code, I noticed that this `mrt` check is not returning NAN. I assume that is the intended behavior to align with `tcppver.out`?

### DIFF
--- a/sgp4/propagation.py
+++ b/sgp4/propagation.py
@@ -1964,6 +1964,8 @@ def sgp4(satrec, tsince, whichconst=None):
                                  ' the satellite has decayed'.format(mrt))
          satrec.error = 6;
 
+         return false, false
+
      return r, v;
 
 """


brandon-rhodes:
Good question! As I finish integrating the new C++ wrapper, I am going to pivot the API toward having folks look at the error code instead of the nan's to determine if there was an error, in response to this issue that I just closed by, it turns out, removing the `return` line of code that used to sit there:

https://github.com/brandon-rhodes/python-sgp4/issues/20
https://github.com/brandon-rhodes/python-sgp4/commit/93b6274b74c8a337e82fa8f8c94da1177326ffe4

Should we add back a comment there to make clear to future readers why we're returning the `r` and `v` even though they're, technically, underground beneath the Earth's surface? :)

dsholes:
Gotcha, that makes sense. Yes, definitely at least a comment. I'm also thinking maybe a warning would be appropriate. That way you don't have to return NAN, or raise an error, but new users who don't read the documentation very carefully will at least know the values are "nonsense".

ckuethe:
The socket module has a few functions with two different error behaviors.. Functions ending with `_ex` throw an exception on error, rather than leaving it to the user to check the returned error code.

brandon-rhodes:
As I update the documentation, I'll include a code snippet for folks to cut and paste showing them how to raise an exception if the error code is nonzero. Unless it winds up being a particularly involved operation, I'm not likely to conclude that the extra code belongs in the library, which ideally will simply be focused on producing the correct numbers.

brandon-rhodes:
@dsholes — Re-reading your original question, I notice for the first time that you are specifically asking about the `tcppver.out` file. I just glanced at it, and am wondering which line specifically you would like me to look at?

brandon-rhodes:
As the new behavior, that puts error checking first as a return value, is now released and documented, I am going to close this for now. We return coordinates where possible, even in the case of beneath-the-Earth's-surface, not just to conform with a test file but generally to get useful results back to people where they exist.

Please comment here if the new docs are inadequate, and we can get them fixed!

PR #45, opened 2020-02-04
Make setup.py forwards compatible with Python 4

Juanlu001:
Discovered using `flake8 --select YTT setup.py` (https://github.com/asottile/flake8-2020)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ ext_modules = []
 
 # It is hard to write C extensions that support both Python 2 and 3, so
 # we opt here to support the acceleration only for Python 3.
-if sys.version_info[0] == 3:
+if sys.version_info[0] >= 3:
     ext_modules.append(Extension(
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
The intention of this `if` statement is to prevent the C code from trying to compile under Python versions other than 3, because Python breaks C-code compatibility with each major version release: you can't expect code written against, say, the Python 2 C API to work against the Python 3 C API.

Therefore, this should not compile against any Python version under than 3, right?

Juanlu001:
You're totally right, I didn't realize the upcoming C API breakage. Sorry for the noise!

brandon-rhodes:
> You're totally right, I didn't realize the upcoming C API breakage. Sorry for the noise!

Alas, as Python 3 demonstrated, backwards compatibility cannot be counted on ahead of time :)

Issue #48, opened 2020-02-05
sgp4init() missing from wrapper.cpp


interplanetarychris:
While Satrec.twoline2rv() allows creating a record from TLEs, CPP-accelerated functionality is missing for creating records directly from element variables.

interplanetarychris:
Addressed by https://github.com/brandon-rhodes/python-sgp4/pull/49

Issue #47, opened 2020-02-04
Allow change of gravity model in Satrec


astrojuanlu:
The new `Satrec` class uses the WGS-72 gravity model:

https://github.com/brandon-rhodes/python-sgp4/blob/10e8076f18bb148d5e2a2434ae20daf9dcfc6fd0/extension/wrapper.cpp#L113

https://github.com/brandon-rhodes/python-sgp4/blob/10e8076f18bb148d5e2a2434ae20daf9dcfc6fd0/sgp4/model.py#L47

There is some discussion in Vallado et al. "Revisiting Spacetrack Report #3: Rev 1" but I haven't fully understood it yet. Do you have any supporting documentation that says that WGS-72 should always be used, or do you think it could be left as a parameter?

brandon-rhodes:
In making that simplification I relied on VI.B. in the paper:

> However, because many operational sites may still have embedded software containing a version of SGP4 using WGS-72, and the fact that the accuracy of the theory would not really be impacted, AFSPC may well have chosen to retain the older set of constants to better maintain interoperability with its internal resources. We use WGS-72 as the default value.

The simplification provided by omitting the parameter was part of making the C wrapper code simple enough that I was able to sit down and get it written and released to users. I figured that folks needing the non-default value could still use the old API, which continues to work with the pure Python code.

Do you have a use case where the difference between them is crucial? I would be happy to look at the difference in output.

Juanlu001:
Thanks for the super quick response!

> I would be happy to look at the difference in output.

I knew you would ask this :) I'll try to produce some numbers to see the actual difference.

Juanlu001:
Here is the difference in output between old API with WGS-84 and 72 and the new one (72):

```
$ cat test_sgp4.py 
import datetime as dt

from sgp4.api import Satrec, jday
from sgp4.io import twoline2rv
from sgp4.earth_gravity import wgs72, wgs84

TLE_LINES = (
    "1 43204U 18015K   18339.11168986  .00000941  00000-0  42148-4 0  9999",
    "2 43204  97.3719 104.7825 0016180 271.1347 174.4597 15.23621941 46156",
)

REF_DATE = dt.datetime(2019, 1, 1)

print(
    twoline2rv(TLE_LINES[0], TLE_LINES[1], wgs84).propagate(
        REF_DATE.year,
        REF_DATE.month,
        REF_DATE.day,
        REF_DATE.hour,
        REF_DATE.minute,
[...]

PR #49, opened 2020-02-05
Add sgp4init to CPP-wrapper and tests

interplanetarychris:
- Adds PyObject `Satrec_sgp4init` modeled after `Satrec_twoline2rv`.
 -Also includes functionality (in wrapper.c and model.py) to automatically set satrec.jdsatepoch and satrec.jdsatepochF as twoline2rv does.
 -Provides access to additional (READONLY) satrec variables for convenience

- Adds PyObject `Satrec_sgp4_tsince` modeled after `Satrec_sgp4` to allow legacy relative timing with new accelerated code.
- As done elsewhere in wrapper.cpp, wgs72 and "improved" ops mode are currently hard-coded
- Extends tests.py to include tests for proper initialization Legacy `sgp4init()` as well as `Satrec.sgp4init()`
 -Combines repeated code for `test_tle_verify()` into subroutine `generate_tle_output()`
 -Includes testing for setting of jdsatrec and jdsatrecF
 -Sets testing of orbital elements as assertAlmostEqual to 12 digits (was getting a fail on the 13th decimal place of nodeo)
 
`Traceback (most recent call last):
  File "tests.py", line 78, in test_satrec_attributes
    self.assertEqual(sat.ndot, 6.96919666594958e-13)
AssertionError: 6.969196665949581e-13 != 6.96919666594958e-13
`

_Seeing sgp4init() speedup of about 25x compared to legacy version, and 10x compared to Cython branch._

_A thorough exploration of OpenMP on Mac OS X Catalina reveals that OpenMP slows everything down by at least 2x._

Closes Issue https://github.com/brandon-rhodes/python-sgp4/issues/48

### DIFF
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -119,7 +119,6 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
 static PyObject *
 Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
 {
-    // char *string1, opsmode[1];
     int satnum;
     double jdSGP4epoch, bstar, ndot, nddot;
     double ecco, argpo, inclo, mo, no_kozai, nodeo;
@@ -127,8 +126,6 @@ Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
     if (!PyArg_ParseTuple(args, "Idddddddddd:sgp4init", &satnum, &jdSGP4epoch, &bstar, &ndot, &nddot,
             &ecco, &argpo, &inclo, &mo, &no_kozai, &nodeo))
         return NULL;
-
-    // strncpy(opsmode, string1, 1);
 
     SatrecObject *self = (SatrecObject*) cls->tp_alloc(cls, 0);
     if (!self)
@@ -169,7 +166,7 @@ static PyMethodDef Satrec_methods[] = {
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
     {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS | METH_CLASS,
-     PyDoc_STR("Initialize the record directly from provided variables.")},
+     PyDoc_STR("Initialize the record from orbital elements.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
      PyDoc_STR("Given minutes since epoch, return position and velocity.")},
     {"_sgp4", (PyCFunction)Satrec__sgp4, METH_VARARGS,
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -95,9 +95,12 @@ static PyObject *
 Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
 {
     char *string1, *string2, line1[130], line2[130];
+    gravconsttype whichconst = wgs72;
     double dummy;
 
-    if (!PyArg_ParseTuple(args, "ss:twoline2rv", &string1, &string2))
+    if (!PyArg_ParseTuple(args,
+            "ss|i:twoline2rv",
+            &string1, &string2, &whichconst))
         return NULL;
 
     // Copy both lines, since twoline2rv() might write to both buffers.
@@ -110,7 +113,7 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
     if (!self)
         return NULL;
 
-    SGP4Funcs::twoline2rv(line1, line2, ' ', ' ', 'i', wgs72,
+    SGP4Funcs::twoline2rv(line1, line2, ' ', ' ', 'i', whichconst,
                           dummy, dummy, dummy, self->satrec);
 
     return (PyObject*) self;
@@ -184,7 +187,7 @@ Satrec__sgp4(PyObject *self, PyObject *args)
 
 static PyMethodDef Satrec_methods[] = {
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
-     PyDoc_STR("Initialize the record from two lines of TLE text.")},
+     PyDoc_STR("Initialize the record from two lines of TLE text and an optional gravity constant.")},
     {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS,
      PyDoc_STR("Initialize the record from orbital elements.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
@@ -342,7 +345,7 @@ Satrec_len(PyObject *self) {
 }
 
 static PySequenceMethods SatrecArray_as_sequence = {
-    .sq_length = Satrec_len,
+    Satrec_len
 };
 
 static PyObject *
@@ -359,10 +362,6 @@ SatrecArray_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
     return type->tp_alloc(type, length);
 }
 
-/* This, and the definition for which it is a forward reference,
-   originally said "static PyTypeObject SatrecArrayType;" but that gave
-   "error: redefinition" during compilation, thus the switch to API. */
-PyAPI_DATA(PyTypeObject) SatrecArrayType;
 
 static int
 SatrecArray_init(SatrecArrayObject *self, PyObject *args, PyObject *kwds)
@@ -407,18 +406,17 @@ static PyMethodDef SatrecArray_methods[] = {
     {NULL, NULL}
 };
 
-PyAPI_DATA(PyTypeObject) SatrecArrayType = {
+static PyTypeObject SatrecArrayType = {
     PyVarObject_HEAD_INIT(NULL, sizeof(elsetrec))
     /* See the module initialization function at the bottom of this file. */
 };
 
 /* The module that ties it all together. */
-
 static PyModuleDef module = {
     PyModuleDef_HEAD_INIT,
-    .m_name = "sgp4.vallado_cpp",
-    .m_doc = "Official C++ SGP4 implementation.",
-    .m_size = -1,
+    "sgp4.vallado_cpp",
+    "Official C++ SGP4 implementation.",
+    -1
 };
 
 PyMODINIT_FUNC
@@ -468,6 +466,11 @@ PyInit_vallado_cpp(void)
         Py_DECREF(m);
         return NULL;
     }
+
+    if (PyModule_AddIntConstant(m, "WGS72", (int)wgs72) ||
+        PyModule_AddIntConstant(m, "WGS72OLD", (int)wgs72old) ||
+        PyModule_AddIntConstant(m, "WGS84", (int)wgs84))
+        return NULL;
 
     return m;
 }
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -117,6 +117,30 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
 }
 
 static PyObject *
+Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
+{
+    // char *string1, opsmode[1];
+    int satnum;
+    double jdSGP4epoch, bstar, ndot, nddot;
+    double ecco, argpo, inclo, mo, no_kozai, nodeo;
+
+    if (!PyArg_ParseTuple(args, "Idddddddddd:sgp4init", &satnum, &jdSGP4epoch, &bstar, &ndot, &nddot,
+            &ecco, &argpo, &inclo, &mo, &no_kozai, &nodeo))
+        return NULL;
+
+    // strncpy(opsmode, string1, 1);
+
+    SatrecObject *self = (SatrecObject*) cls->tp_alloc(cls, 0);
+    if (!self)
+        return NULL;
+
+    SGP4Funcs::sgp4init(wgs72, 'i', satnum, jdSGP4epoch, bstar, ndot, nddot,
+                          ecco, argpo, inclo, mo, no_kozai, nodeo, self->satrec);
+
+    return (PyObject*) self;
+}
+
+static PyObject *
 Satrec_sgp4(PyObject *self, PyObject *args)
 {
     double jd, fr, r[3], v[3];
@@ -144,6 +168,8 @@ Satrec__sgp4(PyObject *self, PyObject *args)
 static PyMethodDef Satrec_methods[] = {
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
+    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS | METH_CLASS,
+     PyDoc_STR("Initialize the record directly from provided variables.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
      PyDoc_STR("Given minutes since epoch, return position and velocity.")},
     {"_sgp4", (PyCFunction)Satrec__sgp4, METH_VARARGS,
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -141,6 +141,21 @@ Satrec_sgp4init(PyObject *self, PyObject *args)
 }
 
 static PyObject *
+Satrec_sgp4_tsince(PyObject *self, PyObject *args)
+{
+    double tsince, r[3], v[3];
+
+    if (!PyArg_ParseTuple(args, "d:sgp4_tsince", &tsince))
+        return NULL;
+    elsetrec &satrec = ((SatrecObject*) self)->satrec;
+    SGP4Funcs::sgp4(satrec, tsince, r, v);
+    if (satrec.error && satrec.error < 6)
+        r[0] = r[1] = r[2] = v[0] = v[1] = v[2] = NAN;
+    return Py_BuildValue("i(f,f,f)(f,f,f)", satrec.error,
+                         r[0], r[1], r[2], v[0], v[1], v[2]);
+}
+
+static PyObject *
 Satrec_sgp4(PyObject *self, PyObject *args)
 {
     double jd, fr, r[3], v[3];
@@ -171,9 +186,11 @@ static PyMethodDef Satrec_methods[] = {
     {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS,
      PyDoc_STR("Initialize the record from orbital elements.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
-     PyDoc_STR("Given minutes since epoch, return position and velocity.")},
+     PyDoc_STR("Given a modified julian date, return position and velocity.")},
     {"_sgp4", (PyCFunction)Satrec__sgp4, METH_VARARGS,
-     PyDoc_STR("Given an array of minutes since epoch, return position and velocity arrays.")},
+     PyDoc_STR("Given an array of modified julian dates, return position and velocity arrays.")},
+    {"sgp4_tince", (PyCFunction)Satrec_sgp4, METH_VARARGS,
+     PyDoc_STR("Given minutes since epoch, return position and velocity.")},
     {NULL, NULL}
 };
 
@@ -181,9 +198,9 @@ static PyMethodDef Satrec_methods[] = {
 
 static PyMemberDef Satrec_members[] = {
     /* Listed in the order they appear in a TLE record. 
-    Expect that direct sgp4init() calls may update most variables below except satnum    */
+    Writeable so that sgp4init() may initialize from an empty Satrec() call */
 
-    {"satnum", T_LONG, O(satnum), READONLY,
+    {"satnum", T_LONG, O(satnum), 0,
      PyDoc_STR("Satellite number, from characters 3-7 of each TLE line.")},
     {"jdsatepoch", T_DOUBLE, O(jdsatepoch), 0,
      PyDoc_STR("Julian date of epoch, day number (see jdsatepochF).")},
@@ -231,8 +248,65 @@ static PyMemberDef Satrec_members[] = {
 
     {"method", T_CHAR, O(method), READONLY,
      PyDoc_STR("Method, either 'n' near earth or 'd' deep space.")},
-    {"error", T_INT, O(method), READONLY,
+    {"error", T_INT, O(method), 0,
      PyDoc_STR("Error code (1-6) documented in sgp4()")},
+    {"a", T_DOUBLE, O(a), 0,
+     PyDoc_STR("semi-major axis")},
+    {"altp", T_DOUBLE, O(altp), 0,
+     PyDoc_STR("altitude of perigee")},
+    {"alta", T_DOUBLE, O(alta), 0,
+     PyDoc_STR("altitude of perigee")},
+
+    /* Single averaged mean elements */
+
+    {"am", T_DOUBLE, O(am), 0,
+     PyDoc_STR("am: Average semi-major axis")},
+    {"em", T_DOUBLE, O(em), 0,
+     PyDoc_STR("em: Average eccentricity")},
+    {"im", T_DOUBLE, O(im), 0,
+     PyDoc_STR("im: Average inclination")},
+    {"Om", T_DOUBLE, O(Om), 0,
+     PyDoc_STR("Om: Average right ascension of ascending node")},
+    {"om", T_DOUBLE, O(om), 0,
+     PyDoc_STR("om: Average argument of perigee")},
+    {"mm", T_DOUBLE, O(mm), 0,
+     PyDoc_STR("mm: Average mean anomaly")},
+    {"nm", T_DOUBLE, O(nm), 0,
+     PyDoc_STR("nm: Average mean motion")},
+
+    /* Gravity-constant dependent values (initialized by sgp4init() */
+
+    {"tumin", T_DOUBLE, O(tumin), 0,
+     PyDoc_STR("minutes in one time unit")},
+    {"mu", T_DOUBLE, O(mu), 0,
+     PyDoc_STR("Earth gravitational parameter")},
+    {"radiusearthkm", T_DOUBLE, O(radiusearthkm), 0,
+     PyDoc_STR("radius of the earth in km")},
+    {"xke", T_DOUBLE, O(xke), 0,
+     PyDoc_STR("reciprocal of tumin")},
+    {"j2", T_DOUBLE, O(j2), 0,
+     PyDoc_STR("un-normalized zonal harmonic j2 value")},
+    {"j3", T_DOUBLE, O(j3), 0,
+     PyDoc_STR("un-normalized zonal harmonic j3 value")},
+    {"j4", T_DOUBLE, O(j4), 0,
+     PyDoc_STR("un-normalized zonal harmonic j4 value")},
+    {"j3oj2", T_DOUBLE, O(j3oj2), 0,
+     PyDoc_STR("j3 divided by j2")},
+
+    /* Other convenience variables (required by propagation.py) */
+    {"t", T_DOUBLE, O(t), 0,
+     PyDoc_STR("Last tsince input to sgp4()")},
+    {"mdot", T_DOUBLE, O(mdot), 0,
+     PyDoc_STR("mean anomaly dot (rate)")},
+    {"mdot", T_DOUBLE, O(mdot), 0,
+     PyDoc_STR("mean anomaly dot (rate)")},
+    {"argpdot", T_DOUBLE, O(argpdot), 0,
+     PyDoc_STR("argument of perigee dot (rate)")},
+    {"nodedot", T_DOUBLE, O(nodedot), 0,
+     PyDoc_STR("right ascension of ascending node dot (rate)")},
+    {"nodecf", T_DOUBLE, O(nodecf), 0,
+     PyDoc_STR("nodecf")},
+
     {NULL}
 };
 
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -92,6 +92,16 @@ _vectorized_sgp4(PyObject *args, elsetrec *raw_satrec_array, int imax)
 /* Details of the "Satrec" satellite object. */
 
 static PyObject *
+Satrec_Satellite(PyTypeObject *cls)
+{
+    SatrecObject *self = (SatrecObject*) cls->tp_alloc(cls, 0);
+    if (!self)
+        return NULL;
+
+    return (PyObject*) self;
+}
+
+static PyObject *
 Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
 {
     char *string1, *string2, line1[130], line2[130];
@@ -116,6 +126,31 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
     return (PyObject*) self;
 }
 
+static bool
+Satrec_sgp4init(PyObject *self, PyObject *args)
+{
+    int satnum;
+    double jdSGP4epoch, bstar, ndot, nddot;
+    double ecco, argpo, inclo, mo, no_kozai, nodeo;
+
+    if (!PyArg_ParseTuple(args, "ldddddddddd:sgp4init", &satnum, &jdSGP4epoch, &bstar, &ndot, &nddot,
+                          &ecco, &argpo, &inclo, &mo, &no_kozai, &nodeo))
+        return false;
+
+    elsetrec &satrec = ((SatrecObject*) self)->satrec;
+
+    SGP4Funcs::sgp4init(wgs72, 'i', satnum, jdSGP4epoch, bstar, ndot, nddot,
+                        ecco, argpo, inclo, mo, no_kozai, nodeo, satrec);
+
+    /* Return true as sgp4init does, satrec.error contains any error codes */
+
+    /* Present state: sgp4init runs fine, but function segmentation faults
+    at the return statement */
+
+    return true;
+}
+
+/*
 static PyObject *
 Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
 {
@@ -136,6 +171,7 @@ Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
 
     return (PyObject*) self;
 }
+*/
 
 static PyObject *
 Satrec_sgp4(PyObject *self, PyObject *args)
@@ -163,9 +199,11 @@ Satrec__sgp4(PyObject *self, PyObject *args)
 }
 
 static PyMethodDef Satrec_methods[] = {
+    {"Satellite", (PyCFunction)Satrec_Satellite, METH_NOARGS | METH_CLASS,
+     PyDoc_STR("Initialize an empty satellite record.")},
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
-    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS | METH_CLASS,
+    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS,
      PyDoc_STR("Initialize the record from orbital elements.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
      PyDoc_STR("Given minutes since epoch, return position and velocity.")},
@@ -177,56 +215,59 @@ static PyMethodDef Satrec_methods[] = {
 #define O(member) offsetof(SatrecObject, satrec.member)
 
 static PyMemberDef Satrec_members[] = {
-    /* Listed in the order they appear in a TLE record. */
+    /* Listed in the order they appear in a TLE record. 
+    Expect that direct sgp4init() calls may update most variables below except satnum    */
 
     {"satnum", T_LONG, O(satnum), READONLY,
      PyDoc_STR("Satellite number, from characters 3-7 of each TLE line.")},
-    {"jdsatepoch", T_DOUBLE, O(jdsatepoch), READONLY,
+    {"jdsatepoch", T_DOUBLE, O(jdsatepoch), 0,
      PyDoc_STR("Julian date of epoch, day number (see jdsatepochF).")},
-    {"jdsatepochF", T_DOUBLE, O(jdsatepochF), READONLY,
+    {"jdsatepochF", T_DOUBLE, O(jdsatepochF), 0,
      PyDoc_STR("Julian date of epoch, fraction of day (see jdsatepoch).")},
-    {"classification", T_CHAR, O(classification), READONLY,
+    {"classification", T_CHAR, O(classification), 0,
      "Usually U=Unclassified, C=Classified, or S=Secret."},
     /* intldesg: inline character array; see Satrec_getset. */
-    {"epochyr", T_INT, O(epochyr), READONLY,
+    {"epochyr", T_INT, O(epochyr), 0,
      PyDoc_STR("Year of this element set's epoch (see epochdays).")},
-    {"epochdays", T_DOUBLE, O(epochdays), READONLY,
+    {"epochdays", T_DOUBLE, O(epochdays), 0,
      PyDoc_STR("Day of the year of this element set's epoch (see epochyr).")},
-    {"ndot", T_DOUBLE, O(ndot), READONLY,
+    {"ndot", T_DOUBLE, O(ndot), 0,
      PyDoc_STR("Ballistic Coefficient in revs/day.")},
-    {"nddot", T_DOUBLE, O(nddot), READONLY,
+    {"nddot", T_DOUBLE, O(nddot), 0,
      PyDoc_STR("Second Derivative of Mean Motion in revs/day^3.")},
-    {"bstar", T_DOUBLE, O(bstar), READONLY,
+    {"bstar", T_DOUBLE, O(bstar), 0,
      PyDoc_STR("Drag Term in inverse Earth radii.")},
-    {"ephtype", T_INT, O(ephtype), READONLY,
+    {"ephtype", T_INT, O(ephtype), 0,
      PyDoc_STR("Ephemeris type (should be 0 in published TLEs).")},
-    {"elnum", T_INT, O(elnum), READONLY,
+    {"elnum", T_INT, O(elnum), 0,
      PyDoc_STR("Element set number.")},
-    {"inclo", T_DOUBLE, O(inclo), READONLY,
+    {"inclo", T_DOUBLE, O(inclo), 0,
      PyDoc_STR("Inclination in radians.")},
-    {"nodeo", T_DOUBLE, O(nodeo), READONLY,
+    {"nodeo", T_DOUBLE, O(nodeo), 0,
      PyDoc_STR("Right ascension of ascending node in radians.")},
-    {"ecco", T_DOUBLE, O(ecco), READONLY,
+    {"ecco", T_DOUBLE, O(ecco), 0,
      PyDoc_STR("Eccentricity.")},
-    {"argpo", T_DOUBLE, O(argpo), READONLY,
+    {"argpo", T_DOUBLE, O(argpo), 0,
      PyDoc_STR("Argument of perigee in radians.")},
-    {"mo", T_DOUBLE, O(mo), READONLY,
+    {"mo", T_DOUBLE, O(mo), 0,
      PyDoc_STR("Mean anomaly in radians.")},
-    {"no_kozai", T_DOUBLE, O(no_kozai), READONLY,
+    {"no_kozai", T_DOUBLE, O(no_kozai), 0,
      PyDoc_STR("Mean motion in radians per minute.")},
-    {"revnum", T_LONG, O(revnum), READONLY,
+    {"revnum", T_LONG, O(revnum), 0,
      PyDoc_STR("Integer revolution number at the epoch.")},
 
     /* For compatibility with the old struct members, also accept the
        plain name "no". */
 
-    {"no", T_DOUBLE, O(no_kozai), READONLY,
+    {"no", T_DOUBLE, O(no_kozai), 0,
      PyDoc_STR("Alias for the more carefully named ``no_kozai``.")},
 
     /* Derived values that do not appear explicitly in the TLE. */
 
     {"method", T_CHAR, O(method), READONLY,
      PyDoc_STR("Method, either 'n' near earth or 'd' deep space.")},
+    {"error", T_INT, O(method), READONLY,
+     PyDoc_STR("Error code (1-6) documented in sgp4()")},
     {NULL}
 };
 
@@ -260,7 +301,7 @@ Satrec_len(PyObject *self) {
 }
 
 static PySequenceMethods SatrecArray_as_sequence = {
-    sq_length : Satrec_len,
+    .sq_length = Satrec_len,
 };
 
 static PyObject *
@@ -334,9 +375,9 @@ PyAPI_DATA(PyTypeObject) SatrecArrayType = {
 
 static PyModuleDef module = {
     PyModuleDef_HEAD_INIT,
-    m_name : "sgp4.vallado_cpp",
-    m_doc : "Official C++ SGP4 implementation.",
-    m_size : -1,
+    .m_name = "sgp4.vallado_cpp",
+    .m_doc = "Official C++ SGP4 implementation.",
+    .m_size = -1,
 };
 
 PyMODINIT_FUNC
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -132,10 +132,12 @@ Satrec_sgp4init(PyObject *self, PyObject *args)
     SGP4Funcs::sgp4init(wgs72, 'i', satnum, jdSGP4epoch, bstar, ndot, nddot,
                         ecco, argpo, inclo, mo, no_kozai, nodeo, satrec);
 
+    /* Populate jdsatepoch and jdsatepochF as SGP4Funcs::twoline2rv does */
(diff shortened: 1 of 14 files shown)


interplanetarychris:
@brandon-rhodes I see the build is failing against the python 2.x-native test against Satrec.sgp4init() -- and that it needs to be added to the model.py Satrec class.

That's a little beyond me, so I'll let you implement that part for python-native fallback!

extension/wrapper.cpp
```diff
@@ -144,6 +168,8 @@ Satrec__sgp4(PyObject *self, PyObject *args)
 static PyMethodDef Satrec_methods[] = {
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
+    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS | METH_CLASS,
+     PyDoc_STR("Initialize the record directly from provided variables.")},
```
brandon-rhodes: Let's be a tiny bit more specific: `Initialize the record from orbital elements.`

sgp4/tests.py
```diff
@@ -85,6 +86,44 @@ def test_satrec_attributes(self):
...
+        ecco = 0.1859667
+        argpo = 5.790416027488515
+        inclo = 0.5980929187319208
+        mo = 0.3373093125574321
+        no_kozai = 0.04722944544077857
+        nodeo = 6.08638547138321
+
+        # Make sure the Satrec ininitalized via sgp4init() has the same attributes.
+        sat = self.build_satrec_sgp4init(satnum, jdsatepoch, bstar, ndot, nddot,
+                          ecco, argpo, inclo, mo, no_kozai, nodeo)
```
brandon-rhodes: Try tweaking the indentation to be some variety of PEP-8 (at least the way GitHub is formatting it, the indent level does not match anything else in the file?). This comment also applies several places below, so try double-checking all the indentation. (If your editor doesn't do it automatically you might check if its settings are off?)

sgp4/tests.py
```diff
@@ -85,6 +86,44 @@ def test_satrec_attributes(self):
...
+        self.assertEqual(sat.satnum, 5)
+        # sgp4init does not back-calculate epochdays - does it need to?
```
brandon-rhodes: Good question! I don't see it used anywhere in the code. For now, feel free to set it to zero if you can't find a convenient call that would back-calculate it.

sgp4/tests.py
```diff
@@ -85,6 +86,44 @@ def test_satrec_attributes(self):
...
+        #self.assertEqual(sat.epochdays, 179.78495062)
+
+        ## sgp4init does not set jdsatepoch internally
```
brandon-rhodes: Feel free to remove this whole comment block — but thanks for leaving it here for me to make a decision about.

sgp4/tests.py
```diff
@@ -165,6 +204,11 @@ class NewSatelliteObjectTests(TestCase, SatelliteObjectTests):
     def build_satrec(self, line1, line2):
         return Satrec.twoline2rv(line1, line2)
 
+    def build_satrec_sgp4init(self, satnum, jdsatepoch, bstar, ndot, nddot,
```
brandon-rhodes: Try expanding the name slightly: `build_satrec_and_call_sgp4init()`.

extension/wrapper.cpp
```diff
@@ -144,6 +168,8 @@ Satrec__sgp4(PyObject *self, PyObject *args)
 static PyMethodDef Satrec_methods[] = {
     {"twoline2rv", (PyCFunction)Satrec_twoline2rv, METH_VARARGS | METH_CLASS,
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
+    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS | METH_CLASS,
```
brandon-rhodes: I'll ask for one small pivot: let’s switch this from a constructor to a plain method. The underlying C++ API makes no restriction on how many times `sgp4init()` is called on a given struct, which allows re-use of an already allocated struct to represent a new (or even a slightly tweaked) satellite. So let’s expose that full power here. Happily, that will not only increase flexibility but should simplify the code a bit.

extension/wrapper.cpp
```diff
@@ -116,6 +116,30 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
     return (PyObject*) self;
 }
 
+static PyObject *
+Satrec_sgp4init(PyTypeObject *cls, PyObject *args)
+{
+    // char *string1, opsmode[1];
```
brandon-rhodes: You can remove these comments when you're done with the code inside.

extension/wrapper.cpp
interplanetarychris: Sorry - This is my first dive into CPython - I've reviewed the docs, but I'm not sure how to implement a "plain method" as opposed to a constructor -- however, agree with you that it would be convenient to call sgp4init on a modified set of variables, repeatedly, and would be glad for the more optimal implementation.  

Could you give this pivot request a go, please?

brandon-rhodes: First, try taking a look at `Satrec__sgp4()` — it’s a plain method taking arguments, exactly the plan you’ll want to use for the init method. My guess is that you’ll just return `None` from the init method since its result is written to the object itself; lots of web resources should be available on how to return that value, as it's a very popular one.

extension/wrapper.cpp
interplanetarychris: Removed.

sgp4/tests.py
interplanetarychris: Updated to be `build_satrec_from_sgp4init`

sgp4/tests.py
interplanetarychris: Removed.

sgp4/tests.py
interplanetarychris: Removed.

sgp4/tests.py
interplanetarychris: Fixed indentation to align with open parenthesis of function on argument linewrap.

extension/wrapper.cpp
interplanetarychris: Taking a look...

brandon-rhodes: Thanks! Given that constructors are actually more complicated than plain methods, my guess is that you'll be able to adapt to the plain-method pattern just fine, given how well you did the constructor.

interplanetarychris:
> To get the tests passing under Python 2.7, I think you'll just need to add a method to `model.py` `Satrec` that calls the `sgp4init()` in `propagation.py`.

I understand that's where the change needs to be made, but per my original commit comment, it wasn't entirely clear (to me) how to add it for fall-back compatibility.

brandon-rhodes:
> I understand that's where the change needs to be made, but per my original commit comment, it wasn't entirely clear (to me) how to add it for fall-back compatibility.

I think the two steps will probably be:

1. Expand `from sgp4.propagation import sgp4` at the top of `model.py` to also import `sgp4init`.
2. Just below the `classmethod` `twoline2rv()`, add a method like:

```
    def sgp4init(self, ... args ...):
        """Initialize this Satrec from raw orbital parameters."""
        return sgp4init(self, ... args ...)
```

If you feel particularly verbose, you could list in the docstring the meanings and units of all the parameters (I think several might be in radians, for example?), but the documentation could always be expanded later.

interplanetarychris:
Current checkins are for the direct-method implementation which currently segmentation-faults.

brandon-rhodes:
Thanks for this additional work! I have tried out the branch and have only two tweaks to suggest:

1. It doesn't matter on most modern laptops, but since `satnum` is an `int`, it should be specified as `i` in the format string. Just in case the sizes are different on a system someone tries to compile it on!
2. Python methods always return a Python object using a pointer ­— so, symmetrically with `Satrec_twoline2rv()` just above it, `Satrec_sgp4init()` must return either `NULL` on error or some object pointer on success; typically with no better value to return, Python docs suggest:

```
    Py_INCREF(Py_None);
    return Py_None;
```

Any misbehavior as you ran the tests was because a boolean value 0 or 1 was being interpreted as a pointer to a Python object and dereferenced, with all the chaos that typically, alas, ensues.

I'm unhappy that my compiler is not pointing out the problem. Shouldn't it emit a warning, at least, that a function with an incompatible return type is being put in a struct that requires another return type? This is why I avoid C in every project I possibly can. :)

interplanetarychris:
> * It doesn't matter on most modern laptops, but since `satnum` is an `int`, it should be specified as `i` in the format string. Just in case the sizes are different on a system someone tries to compile it on!

SGP4.h elsetrec defines satnum as a long int, and a compiler warning is raised if I use a normal int.  If it needs to be "int" SGP4.h should be changed.

interplanetarychris:
>    Py_INCREF(Py_None);
>     return Py_None;

Using this with a Pyobject * return type solves the problem. 

Thanks!

interplanetarychris:
> > > * It doesn't matter on most modern laptops, but since `satnum` is an `int`, it should be specified as `i` in the format string. Just in case the sizes are different on a system someone tries to compile it on!
> > 
> > 
> > SGP4.h elsetrec defines satnum as a long int, and a compiler warning is raised if I use a normal int. If it needs to be "int" SGP4.h should be changed.
> 
> But, if I am reading right, the address you are taking with `&` is not the address of the field in the actual `struct elsetrec`, but the address of your local variable `int satnum;` — if you change it to a `long` then the format string and destination will match, I think?

satnum gets initialized by the elsetrec struct, so it starts out life as a long int in CPP, before getting passed back to pythhon.  Getting passed back into C, its heading into a long int, so it seems like it should always be a long-int.

interplanetarychris:
Note that in order to pass legacy checks, I made a change which removed satrec.whichconst from the struct.  Now if `gravconst` explicitly passed into sgp4(), it will set it with `getgravconst(wgs72)`

The C-code did not have it in the struct, so in this case, the "improved" Legacy code needs to be downgraded to match.

extension/wrapper.cpp
```diff
@@ -116,6 +116,30 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
     return (PyObject*) self;
 }
 
+static PyObject *
+Satrec_sgp4init(PyObject *self, PyObject *args)
+{
+    int satnum;
```
brandon-rhodes: > satnum gets initialized by the elsetrec struct, so it starts out life as a long int … so it seems like it should always be a long-int.

Yes, I agree entirely! My point is that this line makes it a mere `int`, if I am reading correctly, and should instead read `long satnum`?

brandon-rhodes:
1. I'm happy to see the extra `satrec.error` assignment disappearing, since `sgp4init()` itself sets that value to zero. Thanks!
2. Thanks for noticing that `whichconst` is no longer part of Vallado's code.
3. Ping me when the code is again stable and ready for a new review. Thanks!

interplanetarychris:
I'll finish my own review and do a hopefully-final checkin for this particular PR on Saturday.
So far, extremely happy with the speed-ups that the wrapper.cpp has provided my TruSat application.   There are a few other lingering 1.0->2.0 transition items I've noticed, which I might open as issues to allow working each of them to resolution.

brandon-rhodes:
Great! I'm glad to hear I hadn't dropped a ball. No rush, in case your Saturday proves busy; I have plenty of other code to attend to in the meantime.

interplanetarychris:
@brandon-rhodes I think this is ready for review now -- I'll let you resolve the docstring conflicts -- I had upgraded them to a newer method to avoid compilation warnings, but you may have made related changes.

One note on the included is to take a look at forcing wgs72 within `propagation.py` at: https://github.com/brandon-rhodes/python-sgp4/pull/49/files#diff-1c13851eccec30063a7944b50cc30a40R1338-R1339

I believe this was necessary for tests.py to with with the new methods calling non-wrapper functions.

brandon-rhodes:
@interplanetarychris There's still one comment outstanding from the previous round of reviews, I think, if you could take a look at it. Thanks!

extension/wrapper.cpp
interplanetarychris: Fixed with https://github.com/brandon-rhodes/python-sgp4/pull/49/commits/d721c2cee22851ced39502112dfaa016f8e05b28

extension/wrapper.cpp
```diff
@@ -144,66 +183,130 @@ Satrec__sgp4(PyObject *self, PyObject *args)
...
+    {"j3", T_DOUBLE, O(j3), 0,
+     PyDoc_STR("un-normalized zonal harmonic j3 value")},
+    {"j4", T_DOUBLE, O(j4), 0,
+     PyDoc_STR("un-normalized zonal harmonic j4 value")},
+    {"j3oj2", T_DOUBLE, O(j3oj2), 0,
+     PyDoc_STR("j3 divided by j2")},
+
+    /* Other convenience variables (some required by propagation.py) */
+    {"t", T_DOUBLE, O(t), 0,
+     PyDoc_STR("Last tsince input to sgp4()")},
```
brandon-rhodes: There is an unspoken design principle I had in mind when writing this C interface; let's add an Issue to have me add that to some developer docs. The principle was that, to the extent possible, we should take a C++ library that follows the "call-method-and-mutate" approach that's similar to Python's own `list.sort()` and to how PyEphem works, and that instead presents a modern functional interface where object attributes are generally immutable and where methods return information rather than leaving the object with visibly mutated state, more like the modern `sorted()` routine that has replaced `list.sort()` in modern Python code.

Is there a use case for exposing `t`? For each value below, I'm interested in which ones stay static for the lifetime of the object, and which ones change as the satellite is given different `tsince` values.

Hmm.

All of which raises a question I hadn't even thought of until this point: that having `sgp4init()` as a method — like I asked when you'd so nicely wrapped it up as a constructor originally — that mutates the satellite object over its lifetime into a series of different satellite orbits, a new one each time it's called, kind of contradicts how I'm asking for attributes to be handled where they'd ideally be unchanging and read-only.

I think the big question that will resolve my inconsistency here is: what are all the use cases you are working to cover with the addition of sgp4init() and the exposure of additional attributes? Happily, our need to add those use cases to the documentation means that you won't be wasting effort to write up the use cases; that work will feed directly into the big docstring that gets sent up to PyPI.

Feel free to not write the documentation text yourself, as I'm happy to take on that burden. If you'd like, just add a little `newdocs.md` file at the root of this PR that says things like `For reason <X> folks in situation <Y> need the grav parameters to do <Z>, so:` and then shows a small complete script that demonstrates to users how to accomplish it. I can then go through them, transfer them to the main docstring, and write bigger explanations once this PR is merged.

And, as I read through them, I'll understand the use cases covered by these new features, and will be in a better position to decide on whether I made a mistake by suggesting that `sgp4init()` should be a plain method.

I wonder if my thought was "only advanced users would ever call `sgp4init()` so why not have it update the state directly?"

extension/wrapper.cpp
```diff
@@ -237,7 +340,7 @@ Satrec_len(PyObject *self) {
 }
 
 static PySequenceMethods SatrecArray_as_sequence = {
-    sq_length : Satrec_len,
+    .sq_length = Satrec_len,
```
brandon-rhodes: Once the PR is nearly ready, I'll plan on trying it out on the `release-test` branch to see what Mac OS X and Windows think of this syntax; I can't remember if they both like them or not.

sgp4/functions.py
```diff
@@ -8,7 +8,7 @@
 def jday(year, mon, day, hr, minute, sec):
      """Return two floats that, when added, produce the specified Julian date.
 
-     The first float returned gives the date, while the second float
+     The first float returned (jd) gives the date, while the second float (fr)
```
brandon-rhodes: Nice: I love the idea of, even in the docstring (and not just in the main docs), telling users good names for the two return values.

But I'm not sure we should do it mid-paragraph? Unless we decide to return a namedtuple, these won't be names that programmers can use. If they try something like:
```
result = jday(...)
print(result.jd)
```
— then they'll get a `NameError`. So I suspect that advertising the names might be confusing, since by default Python tuples arrive with no fixed names for their values.

**However,** I do love suggesting good names to developers, it's a great practice, it leads to more readable user code and more consistent code, so I completely agree with your goal here. So here's another way to go about it: add to the docstring a little sample line of code showing how to call it:
```
You can call this like:

jd, fr = jday(...)
```
That way, we both teach users about the best names to use, make their code more consistent, but also make it clear _how_ the values are going to get those names in their code: not because we are setting those names, but because the users themselves will set those names voluntarily through tuple unpacking. Otherwise, novice Python programmers might not know exactly how to give those two names to the two return values.

sgp4/model.py
```diff
@@ -48,6 +48,15 @@ def twoline2rv(cls, line1, line2):
         self.epochyr %= 100  # undo my non-standard 4-digit year
         return self
 
+    @classmethod
+    def Satellite(cls):
```
brandon-rhodes: This classmethod is slightly surprising. Could you explain it?

sgp4/propagation.py
```diff
@@ -1129,8 +1129,8 @@ def _dspace(
 """
 
 def _initl(
-       # not needeed. included in satrec if needed later
-       # satn,
+       # not needeed. included in satrec if needed later
```
brandon-rhodes: Is this extra whitespace intentional, or the result of a merge or something like that? (Also take a look at the several times it happened further down.)

sgp4/propagation.py
```diff
@@ -1335,8 +1335,8 @@ def sgp4init(
      #  ------------------------ earth constants -----------------------
      #  sgp4fix identify constants and allow alternate values
      #  this is now the only call for the constants
-     (satrec.tumin, satrec.mu, satrec.radiusearthkm, satrec.xke,
-       satrec.j2, satrec.j3, satrec.j4, satrec.j3oj2) = whichconst;
+     (satrec.tumin, satrec.mu, satrec.radiusearthkm, satrec.xke, 
+       satrec.j2, satrec.j3, satrec.j4, satrec.j3oj2) = getgravconst('wgs72');
```
brandon-rhodes: This looks like it breaks the previous API we still need to support, which used the gravity parameters passed in. Here you are still accepting the parameter but not using it?

sgp4/propagation.py
```diff
@@ -1680,11 +1680,9 @@ def sgp4init(
   ----------------------------------------------------------------------------*/
 """
 
-def sgp4(satrec, tsince, whichconst=None):
+def sgp4(satrec, tsince):
```
brandon-rhodes: This also looks like it would break old code. Can we revert and restore compatibility?

sgp4/tests.py
```diff
@@ -84,6 +85,33 @@ def test_satrec_attributes(self):
         self.assertEqual(sat.mo, 0.3373093125574321)
         self.assertEqual(sat.no_kozai, 0.04722944544077857)
 
+    def test_satrec_sgp4init_attributes(self):
```
brandon-rhodes: Nice! I like that you've added a test that does a simple round-trip of the new method.

1. Is there any way to test if the epoch got set too, or is that difficult to test the way the object is designed?
2. This is minor, but: could the attributes go in the same order in the list of attributes and the list of checks? Otherwise it requires a bit of visual searching to see if the values match up in the test.

sgp4/tests.py
```diff
@@ -164,6 +192,17 @@ class NewSatelliteObjectTests(TestCase, SatelliteObjectTests):
     def build_satrec(self, line1, line2):
         return Satrec.twoline2rv(line1, line2)
 
+    def build_satrec_from_sgp4init(self, satnum, jdsatepoch, bstar, ndot, nddot,
```
brandon-rhodes: Nice, thanks for adding these helper methods! Are you planning to call them from a test? Or, will it be a subsequent pull request that adds some tests that use them?

extension/wrapper.cpp
```diff
@@ -116,6 +116,45 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
...
+        return NULL;
+
+    elsetrec &satrec = ((SatrecObject*) self)->satrec;
+
+    SGP4Funcs::sgp4init(wgs72, 'i', satnum, jdSGP4epoch, bstar, ndot, nddot,
+                        ecco, argpo, inclo, mo, no_kozai, nodeo, satrec);
+
+    /* Return true as sgp4init does, satrec.error contains any error codes */
+
+    /* Present state: sgp4init runs fine, but function segmentation faults
```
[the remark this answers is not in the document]
interplanetarychris: No longer true, removed comment.

extension/wrapper.cpp
```diff
@@ -116,6 +116,45 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
...
+{
+    double tsince, r[3], v[3];
+
+    if (!PyArg_ParseTuple(args, "d:sgp4_tsince", &tsince))
+        return NULL;
+    elsetrec &satrec = ((SatrecObject*) self)->satrec;
+    SGP4Funcs::sgp4(satrec, tsince, r, v);
+    if (satrec.error && satrec.error < 6)
+        r[0] = r[1] = r[2] = v[0] = v[1] = v[2] = NAN;
+    return Py_BuildValue("i(f,f,f)(f,f,f)", satrec.error,
```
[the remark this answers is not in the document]
interplanetarychris: Fixed. This was part of exploring returning something other than a tuple.

I'd still love for this to return a C-array (or python array.array) without touching numpy, but have not figured that out yet.  For these 3d vectors, raw routines with python array.array are significantly (3-8x) faster than numpy routines.

extension/wrapper.cpp
```diff
@@ -144,66 +183,130 @@ Satrec__sgp4(PyObject *self, PyObject *args)
...
      PyDoc_STR("Initialize the record from two lines of TLE text.")},
+    {"sgp4init", (PyCFunction)Satrec_sgp4init, METH_VARARGS,
+     PyDoc_STR("Initialize the record from orbital elements.")},
     {"sgp4", (PyCFunction)Satrec_sgp4, METH_VARARGS,
-     PyDoc_STR("Given minutes since epoch, return position and velocity.")},
+     PyDoc_STR("Given a modified julian date, return position and velocity.")},
     {"_sgp4", (PyCFunction)Satrec__sgp4, METH_VARARGS,
-     PyDoc_STR("Given an array of minutes since epoch, return position and velocity arrays.")},
+     PyDoc_STR("Given an array of modified julian dates, return position and velocity arrays.")},
+    {"sgp4_tince", (PyCFunction)Satrec_sgp4, METH_VARARGS,
```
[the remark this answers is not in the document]
interplanetarychris: Turns out I haven't actually been calling this function in my own code, after internally adjusting for the array methods.  It already was present in tests.py [invoke_satrec_tsince()](https://github.com/interplanetarychris/python-sgp4/blob/d721c2cee22851ced39502112dfaa016f8e05b28/sgp4/tests.py#L202-L204) but not included in the broader test coverage.  Will have to study this a bit to figure out how to make it work for both New and Legacy satellite objects.

extension/wrapper.cpp
```diff
@@ -144,66 +183,130 @@ Satrec__sgp4(PyObject *self, PyObject *args)
...
+     PyDoc_STR("Given minutes since epoch, return position and velocity.")},
     {NULL, NULL}
 };
 
 #define O(member) offsetof(SatrecObject, satrec.member)
 
 static PyMemberDef Satrec_members[] = {
-    /* Listed in the order they appear in a TLE record. */
+    /* Listed in the order they appear in a TLE record. 
+    Writeable so that sgp4init() may initialize from an empty Satrec() call */
```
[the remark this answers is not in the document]
interplanetarychris: `AttributeError: readonly attribute`

extension/wrapper.cpp
```diff
@@ -144,66 +183,130 @@ Satrec__sgp4(PyObject *self, PyObject *args)
...
 
-    {"no", T_DOUBLE, O(no_kozai), READONLY,
+    {"no", T_DOUBLE, O(no_kozai), 0,
      PyDoc_STR("Alias for the more carefully named ``no_kozai``.")},
 
     /* Derived values that do not appear explicitly in the TLE. */
 
     {"method", T_CHAR, O(method), READONLY,
      PyDoc_STR("Method, either 'n' near earth or 'd' deep space.")},
+    {"error", T_INT, O(method), 0,
```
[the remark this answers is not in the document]
interplanetarychris: It's impossible to initialize an object via sgp4init without being able to write to the variables directly.

That covers the following which are derived from the TLE in twoline2rv:
1. satrec.satnum
1. satrec.ndot
1. satrec.nddot
1. satrec.bstar
1. satrec.ephtype
1. satrec.elnum
1. satrec.inclo
1. satrec.nodeo
1. satrec.ecco
1. satrec.argpo
1. satrec.mo
1. satrec.no_kozai
1. satrec.revnum
1. satrec.jdsatepoch
1. satrec.jdsatepochF

If you would like to reinitialize a satrec based on a propagated record (say for example, to read the averaged mean elements and change the epoch of a satrec), you need to be able to read the following satrec variables, and update their equivalents above:
1. sat.im
1. sat.Om
1. sat.em
1. sat.om
1. sat.mm
1. sat.no_kozai (already covered above)

I can't think of failsafe way to allow writing them, but warn the user what they're changing.  SGP4.CPP doesn't internally keep track of what the variables *were*.  Maybe a "write-enable" flag that could be efficiently parsed, while keeping them declared as writeable?

extension/wrapper.cpp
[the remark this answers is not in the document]
brandon-rhodes: Thanks!

extension/wrapper.cpp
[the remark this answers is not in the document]
brandon-rhodes: Can you call the array-based method with a 1-element time array? That should give you NumPy arrays back without the expense of intermediate Python representations. Let me know if you can't get that to work, and I can suggest sample code! (And will them update the docs to make the approach clearer.)

extension/wrapper.cpp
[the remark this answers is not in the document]
brandon-rhodes: Thanks! Let me know what you figure out.

extension/wrapper.cpp
interplanetarychris: Documentation is nice, more especially for advanced use cases.

In essence, my application is repeatedly reinitializing a satrec instance with tweaked elements to evaluate its affect on RMS errors when compared with observations.  Being able to work with the raw orbital elements, rather than having to re-construct a TLE to then re-import has significant ease-of-use as well as performance considerations.

sgp4/functions.py
interplanetarychris: You've already called them jd, fr -- I was making the documentation more explicit than requiring the user to mentally equate first = jd, second = jr

sgp4/model.py
interplanetarychris: I can't - just following your example.

sgp4/propagation.py
interplanetarychris: Don't know why its there.

sgp4/propagation.py
interplanetarychris: This might need to be re-written.  satrec.whichconst doesn't exist in the new satellite object.  We could accept it, then throw it out.  It's part of the larger problem in the new object of hard-coding wgs72 which is highlighted in issue https://github.com/brandon-rhodes/python-sgp4/issues/47

In this case, the backward compatibility appears to be compromised.

sgp4/propagation.py
interplanetarychris: See previous comment.

sgp4/tests.py
interplanetarychris: That was apparently my oversight, and not completely understanding the unittests.  Will need to add methods to call them, and working on a PR that works for both new and legacy methods.

extension/wrapper.cpp
brandon-rhodes: Thanks for the details! How often are you re-submitting the whole slate of elements, versus tweaking just one element and leaving the others static (since the latter case would be a very strong case for a writable element, since calling `sgp4init()` all over again would uselessly set all the elements over again to the values they already have)?

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: Perhaps a poor example.  Take a case were you're updating the elements to adjust an epoch.

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: I was more speaking of the r,v vectors as r[3] and v[3] C-arrays

sgp4/propagation.py
brandon-rhodes: Could you show the exception that was raised when this class was left as-is? I'm not sure I've yet caught the problem you're trying to solve through this change.

sgp4/tests.py
brandon-rhodes: Roger! Here are the docs that I used in writing this test suite, in case they're helpful:

https://docs.python.org/3.8/library/unittest.html#basic-example

sgp4/propagation.py
brandon-rhodes: Roger. I'll try adding `pyflakes` to the project's CI, so that it catches trailing whitespace as part of the build, so that this can be caught automatically next time.

interplanetarychris: I've removed the errant whitespace in an upcoming commit.

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: I'm going to explore changes to my own code which would require that all changes to Satrec() be made through the sgp4init() call.  This has the downside of sometimes requiring intermediate storage of update variables, but at the benefit of guaranteed consistency as you highlight.

My changes would still expose a few more variables (as read-only) for the reasons described above.

brandon-rhodes:
@interplanetarychris — For a second time this afternoon, GitHub's alerts have brought me here, yet I can't find any events on the PR marked as having happened in April so that I could take a look at them. Do you know what might be triggering the alert?

interplanetarychris:
> @interplanetarychris — For a second time this afternoon, GitHub's alerts have brought me here, yet I can't find any events on the PR marked as having happened in April so that I could take a look at them. Do you know what might be triggering the alert?

I've "resolved" a few conversations, added comments to others, and committed two updates to my own branch.  Hoping to have a revised PR to submit for your review in the next few days.

brandon-rhodes:
> I've "resolved" a few conversations, added comments to others, and committed two updates to my own branch. Hoping to have a revised PR to submit for your review in the next few days.

Ah, that makes sense! The UI isn't highlighting the "resolved" part so I feared that there might be questions or comments that I wasn't seeing and would leave dangling. Carry on!

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: @brandon-rhodes I've found an acceptable implementation of `sgp4init()` which leaves the variables as read-only, exposes additional (read-only) variables for convenience with **one key exception** - the epoch-related variables:
1. satrec.jdsatepoch
1. satrec.jdsatepochF
1. satrec.epochyr
1. satrec.epochdays

When a `Satrec()` is initialized with twoline2rv, these variables are set/derived from the epoch data in the TLE - giving a convenient internal reference for what epoch the `Satrec` is associated with. 

Vallado's `sgp4init()` offers no equivalent, and no internal functions with which to call in order to derive the variables - the SGP4epoch -> MJD epoch is easy enough, but converting either of those to a year and DOY are not possible, as there is not an inverse of `days2mdhms()`

Since we're treating the Vallado C-code as "read-only", I would propose allowing the above 4 variables to be writable such that `sgp4init()` users can create a workflow and structure equivalent to `twoline2rv()` calls (if they care to).  This would appear to be safe, as these variables are only set and referenced by `twoline2rv()`, but not used by any other SGP4 function.

`twoline2rv()` would automatically overwrite anything with variables consistent with the provided TLE, and anyone who is using them via `sgp4init()` would have had to take the explicit action to set them as well.

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: Tried calling the array-based method with a 1-element time array to get a numpy array "for free", but it comes at a penalty of being about 3x slower.  The crossover between multiple calls of the single method vs. the array method appears to be around ~5 elements in my environment.  Given that I'm almost always dealing with 3 or more time-elements, it might be worth favoring the array method anyway.  I'll explore this a little more thoroughly with a closer evaluation of the performance hit of the totality of the array-setup overhead.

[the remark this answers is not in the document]
interplanetarychris: FYI - Finished implementing an all array (_sgp4) implementation in my own code.  In the course of doing that, also explored the openmp "parallel for" options.  I tried various optimizations of your default code, but generally saw that it was 2.5x slower for small numbers (10) and 60% slower for (100) elements on an 18-core machine with openmp from `brew install llvm libomp`.  At ten entries, the "no openmp" version is about 20% faster, but don't have test data to compare at 100 element.

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: Resolved this by having model.Satrec.sgp4init() and wrapper.c:sgp4init() set satrec.jdsatepoch and satrec.jdsatepochF so it will work the same everywhere in the V2 code.

Let epochyr and epochdays unset (even though twoline2rv sets them), as there's no existing code within the code base to get epochdays, and it's not used by any of the other functions.

sgp4/tests.py
interplanetarychris: (1) Done and (2) done.

sgp4/tests.py
interplanetarychris: Figured out the tests.  Notes in the upcoming commit.

extension/wrapper.cpp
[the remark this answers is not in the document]
interplanetarychris: Fixed in the upcoming commit.

sgp4/propagation.py
interplanetarychris: Reverted to original code, with recent updates to whichconst.

sgp4/propagation.py
interplanetarychris: Reverted to original code with recent changes to whichconst.

interplanetarychris:
@brandon-rhodes All changes incorporated, tests passing, and rebased.

brandon-rhodes:
> @brandon-rhodes All changes incorporated, tests passing, and rebased.

Thanks! I'll take a look early this week as soon as there's a gap in my other duties.

brandon-rhodes:
Thanks, I've just read over the diff! Three meta-questions before I dive into line-level comments:

1. Has there been a loss of precision in building the objects? I see previously solid tests being reduced to "almost equal" tests, implying that the floats are being degraded in accuracy as they're copied from Python to C++. I'd prefer to fix that problem rather than weaken the tests.
2. Wow, that's a lot of extra attributes ­— some of which, like the gravity constants, aren't even properties of any individual satellite. Could you remind me what our guideline is for what to expose? I'd like to go ahead and document it, so that we know we have a bright line for the future explaining what we chose to turn into a Python-level attribute. (Thanks for making them all read-only, though, I think that will be a big help for users!)
3. What if in the test suite you created a new class, parallel to the classes `NewSatelliteObjectTests` and `LegacySatelliteObjectTests`, called maybe `InitSatelliteObjectTests` whose `build_satrec()` calls your new function? The intention was roughly that for every way to build a satellite, there would be a subclass that specialized `build_satrec()` and then thanks to the magic of inheritance, that satellite would have all its tests run.
4. That previous point might alleviate the need for `generate_tle_output()` as a generator — but if you think the code is prettier with it as a generator, we can keep it, it does reduce a level of indentation!

interplanetarychris:
> On Apr 20, 2020, at 1:35 PM, Brandon Rhodes <notifications@github.com> wrote:
> 
> 
> Thanks, I've just read over the diff! Three meta-questions before I dive into line-level comments:
> 
> Has there been a loss of precision in building the objects? I see previously solid tests being reduced to "almost equal" tests, implying that the floats are being degraded in accuracy as they're copied from Python to C++. I'd prefer to fix that problem rather than weaken the tests.
I just saw one problem (mentioned in the updated description for the PR).  If you want to put it back to “AssertEqual”’s, I’m fine with that.

> Wow, that's a lot of extra attributes ­— some of which, like the gravity constants, aren't even properties of any individual satellite. Could you remind me what our guideline is for what to expose? I'd like to go ahead and document it, so that we know we have a bright line for the future explaining what we chose to turn into a Python-level attribute. (Thanks for making them all read-only, though, I think that will be a big help for users!)
They are all things that I’ve found on one occasion or another are useful to be able to read. 

For example, the gravity parameters make it easy to derive other values without having to create other code or variables to manage what gravity constant was used (its already in the satrec… use it!)

I don’t believe it costs anything, and now that they’re read-only, should be without risk.

The C-comments in wrapper.c were my initial rationale for including them.
> What if in the test suite you created a new class, parallel to the classes NewSatelliteObjectTests and LegacySatelliteObjectTests, called maybe InitSatelliteObjectTests whose build_satrec() calls your new function? The intention was roughly that for every way to build a satellite, there would be a subclass that specialized build_satrec() and then thanks to the magic of inheritance, that satellite would have all its tests run.
That sounds workable.  Sub-classing classes is just a little beyond my level of comfort, but I believe I understand what you’re describing - and it would allow the elimination of redundant code.
> That previous point might alleviate the need for generate_tle_output() as a generator — but if you think the code is prettier with it as a generator, we can keep it, it does reduce a level of indentation!
Possibly - I didn’t like the “Raise Error” calls in those, as they’re not very “unit testy” — you could instead assign those outputs to a ‘msg’ variable which AssertTrue could then output to the user — it would show up as a “FAIL” instead of an “ERROR” which is more reflective of what’s happening.  Unfortunately, making it a generator loses access to self.*

interplanetarychris:
A further "bright line" in regards to satrec.* attributes: You could include something like the following statement in the documentation:

_Variables computed by SGP4 and stored internally in Satrec() are made available (readonly) where they provide context (such as the gravitation-related constants used to initialize the satellite), or eliminate the need to repeat calculations already performed by sgp4 (e.g. `satrec.a` for semi-major axis, `satrec.altp` for perigee, etc.)_

brandon-rhodes:
> I just saw one problem (mentioned in the updated description for the PR).  If you want to put it back to “AssertEqual”’s, I’m fine with that.

Thanks! Put everything back that you can, and I'll look at the remaining cases you have to leave where the matches aren't perfect, once I've merged it.

> They are all things that I’ve found on one occasion or another are useful to be able to read.

Okay, I just wanted to double-check! Let's keep them, then. (There's one duplicate, I think, but I can easily enough remove it later.)

> Sub-classing classes is just a little beyond my level of comfort…

Yes, I wish that Python's unittest had used something other than classes. Let me know if you have any questions about the subclass pattern I'm using — it might just be cut-and-paste to get it running against your new code.

> I didn’t like the “Raise Error” calls in those, as they’re not very “unit testy”…

Good point! The `SatelliteObjectTests` are definitely integration tests, as they test a satellite end-to-end against real application data. Would adding the word "integration" in there somewhere help clarify that not everything in `tests.py` is at the unit testing level? I guess at the moment it's a bit of a hodge-podge without a clear point at which unit tests end and integration tests start.

And thanks for describing the line! I'll integrate that once this is landed.

interplanetarychris:
@brandon-rhodes:
- I've eliminated the redundant `satrec` element (thanks for spotting that!)
- I've returned the attribute tests to be "AssertEqual" and none are currently failing for the code I've contributed (although `test_three_gravity_models` has some errors)

I'm not sure I follow your logic on creating a new TestObject for the ways that a satellite is initialized.  Currently it is split into "legacy code" (no CPP wrapper) and the new code.  If I were to add methods related to how sats are initialized/delta-ed, I believe I would need to add "sgp4init" as well as sgp4_tsince" but upon my evaluation of the code, I'm not sure how that would make the tests cleaner...

brandon-rhodes:
> If I were to add methods related to how sats are initialized/delta-ed, I believe I would need to add "sgp4init" as well as sgp4_tsince" but upon my evaluation of the code, I'm not sure how that would make the tests cleaner...

So, looking at the situation again now that I've been outside and taken a walk, you're right that another subclass doesn't make since for sgp4init, because you can't pass sgp4init a TLE — so the whole crazy routine that tries to output that old legacy TLE test-file format isn't relevant.

But — in that case, I'm confused about why your diff was moving it around. Let me go re-read.

brandon-rhodes:
It makes more sense each time I re-read it. From my most recent read-through:

1. Your tests can, happily, be freed from the `SatelliteObjectTests`. That class is only for tests that need to be exercised twice, once for the old-style API and once for the new. But as this `sgp4init()` method is only being added to the new classes in `model.py`, it doesn't need to happen twice, once for the legacy API and once for the new. You can just add  a single new test to `NewSatelliteObjectTests` that doesn't participate at all in the other methods and instead just builds a satellite and then exercises it to make sure it got initialized correctly.
2. So I think `build_satrec_from_sgp4init()` can go away?
3. Hmm. `test_tle_verify_tsince()` is an interesting case. It does look indeed like a whole new way to generate all of our satellite results. If you want to test it against every single satellite every, give it its own new subclass `class NewSatelliteObjectTsinceTests(NewSatelliteObjectTests):` where the only difference is that you override `invoke_satrec()` to be much simpler and merely pass through `tsince`. Which I think means `invoke_satrec_tsince()` and `test_tle_verify_tsince()` go away — which I think is the simplification I was sensing earlier but not describing well.
4. So we can either leave `generate_tle_output()` in place or put it back, at your option.

Let me know if that makes more sense and is less vague than my last attempt, and thanks for your patience! It's been a while since I've read through the test suite.

interplanetarychris:
It keeps getting deeper - maybe time for a call to discuss?  A few comments below.

> 1. Your tests can, happily, be freed from the `SatelliteObjectTests`. That class is only for tests that need to be exercised twice, once for the old-style API and once for the new. But as this `sgp4init()` method is only being added to the new classes in `model.py`, it doesn't need to happen twice, once for the legacy API and once for the new. You can just add  a single new test to `NewSatelliteObjectTests` that doesn't participate at all in the other methods and instead just builds a satellite and then exercises it to make sure it got initialized correctly.
> 2. So I think `build_satrec_from_sgp4init()` can go away?

From a test coverage point, perhaps not - previously, the sgp4init() function (the original one from V1) was not being tested at all, except indirectly by twoline2rv calling it.  Including the function allows greater code-test coverage for both the old and new models.

> 3. Hmm. `test_tle_verify_tsince()` is an interesting case. It does look indeed like a whole new way to generate all of our satellite results. If you want to test it against every single satellite every, give it its own new subclass `class NewSatelliteObjectTsinceTests(NewSatelliteObjectTests):` where the only difference is that you override `invoke_satrec()` to be much simpler and merely pass through `tsince`. Which I think means `invoke_satrec_tsince()` and `test_tle_verify_tsince()` go away — which I think is the simplification I was sensing earlier but not describing well.

Yes - I think I follow.  While it's not necessary to test it against every case, a new subclass might be the simplest way to code it (rather than making a short version just to test that the variables are passed in appropriately).

> 4. So we can either leave `generate_tle_output()` in place or put it back, at your option.
I think put it back.

When studying this further, I see another problem with test coverage for V2 - it doesn't appear that the V2 object methods from model.* get tested, just the ones from `wrapper.py/wrapper.cpp`.  I'm willing to bet my new entries in model.py don't work appropriately, as the testing does not cover them from that location.

I'll attempt another iteration.

interplanetarychris:
c7543fe covers your most-recently suggested changes, but before I attempt any changes to expand test coverage to include fall-back functions in model.py

interplanetarychris:
I believe a1d9e3b facilitates unit tests against the "slow" V2 object code in `model.py`

I had bugs in my own code (from not testing) and made other changes necessary in `model.py` for the other function tests to pass.

brandon-rhodes:
> previously, the sgp4init() function (the original one from V1) was not being tested at all

Yes, that's true — since it was never documented, I was content to leave it secret and not officially supported. The other routines that use it would always make sure it kept working, of course, but until you suggested a use case for calling it directly, I had not considered documenting it as an actual feature of the library for the use of other Python projects. Since the only documentation for it will be against the new API and not the old one (which is no longer documented), it would be symmetric with the documentation to keep the old one uncalled in the tests.

But — if you've written a test for it, I won't complain, as there's the possibility that the old API's version of the routine will be in use somewhere even though it was undocumented.

I test both the C++ and Python code by running `tox`. See the `tox.ini` in the repository for details. I'll go add that to the documentation, for folks making improvements that touch both versions of the code.

brandon-rhodes:
I'm done reading! Thanks for this most recent update. I think the code is in good enough shape to merge. I'll next go in and completely re-do how testing works, so that there are no subclasses. That means of organization worked okay when the test suite was small, but now prevents me from easily reading what's going on. I'll see if I can get it improved.

interplanetarychris:
@brandon-rhodes When does the update get pushed over to pypi? `sgp4==2.6`? (or will you rev to 2.7)

brandon-rhodes:
@interplanetarychris — The routine now, thanks to the new gravity constants, accepts exactly the same argument list as its C++ counterpart (except the last since the object argument is implicit), and I've updated the docs in `__init__.py` to tell users about the routine and to document its arguments. I plan to release later today, but if you want to take a look first feel free to offer feedback.

interplanetarychris:
@brandon-rhodes 
For sanity, I would recommend using `jdSGP4epoch` everywhere where the 1950 epoch (in the code, usually `epoch`) is used - it helps (as a developer) from going insane remembering which is which.

Out of curiosity, why subtract 0.5 from jdsatepoch when splitting from `jd` - > `jd,jdF` ?  (And if so, why not always do it?)

https://github.com/brandon-rhodes/python-sgp4/blob/225b830d8e4213295a813beef36467211ade4545/sgp4/model.py#L56

You mention whichconst options `a` vs `i` in the docs, but V2 Satrec hard-codes `i` within `twoline2rv` and `sgp4init` -- it looks simple enough to pass through the argument, and perhaps you forgot to propagate the change everywhere.
 
...similarly, `sgp4init` still hardcodes `whichconst=wgs72` and probably doesn't need to anymore.
'
Last, as `SGP4.cpp` doesn't store `satrec.whichconst` it might be worth documenting the loss of the internal variable from V1->V2.

brandon-rhodes:
> For sanity, I would recommend using `jdSGP4epoch` everywhere where the 1950 epoch (in the code, usually `epoch`) is used - it helps (as a developer) from going insane remembering which is which.

Alas, for me sanity ran in the other direction: since `epoch` is the name in Vallado’s reference implementation, I kept having to double-check what a `jdSGP4epoch` is. But I’ll have a chance to reconsider the next time I return to the code, and have to recognize all the variable names all over again! If I find `epoch` opaque when I return then I’ll keep your recommendation in mind.

> Out of curiosity, why subtract 0.5 from jdsatepoch when splitting from `jd` - > `jd,jdF` ?

I wanted the JD to end with 0.5, per astronomy convention. Since `divmod(x, 1.0)` produces a number that ends in .0, I need to bump 0.5 sidesways beforehand and then back the other way afterward to produce JDs that look normal to astronomers, like "2458826.5".

> (And if so, why not always do it?)

That's a great idea, I agree with you! If you'll point me at places where we produce a JD that ends in .0, I'll go get them fixed up.

> You mention whichconst options `a` vs `i` in the docs, but V2 Satrec hard-codes `i` within `twoline2rv` and `sgp4init` -- it looks simple enough to pass through the argument, and perhaps you forgot to propagate the change everywhere. ...similarly, `sgp4init` still hardcodes `whichconst=wgs72` and probably doesn't need to anymore.

Alas! I went to all the trouble to add those arguments, then I forgot to update the C++ call to actually use them. THANK YOU for pointing that out! I'll go write a test first that shows both errors, then fix them, before doing the release.

> Last, as `SGP4.cpp` doesn't store `satrec.whichconst` it might be worth documenting the loss of the internal variable from V1->V2.

I'm not sure where I'd put that in the docs. Do we have a section on migrating to the new API anywhere? I'm trying to remember.

interplanetarychris:
> > Out of curiosity, why subtract 0.5 from jdsatepoch when splitting from `jd` - > `jd,jdF` ?
> 
> I wanted the JD to end with 0.5, per astronomy convention. Since `divmod(x, 1.0)` produces a number that ends in .0, I need to bump 0.5 sidesways beforehand and then back the other way afterward to produce JDs that look normal to astronomers, like "2458826.5".
> 
> > (And if so, why not always do it?)
> 
> That's a great idea, I agree with you! If you'll point me at places where we produce a JD that ends in .0, I'll go get them fixed up.
> 
> >


https://github.com/brandon-rhodes/python-sgp4/blob/091b8ee08c8c8b5a8d2168731918af9afdcc3f21/sgp4/model.py#L67-L68

brandon-rhodes:
Given an input like `123.456`, that code on my machine produces `jdsatepoch == 2433404.5` and 'jdsatepochF == 0.456`, which are the values we want, I think?

interplanetarychris:
I'm told that's what astronomers are used to seeing.

brandon-rhodes:
As long as we've chosen a popular convention and stick with it through the codebase, I think I'm happy. Let me know if you run across any other potential inconsistencies, or any spots where the values create any problem for particular use cases that aren't yet in the tests.

brandon-rhodes:
@interplanetarychris — After an exhausting day and a half of tracking down small inconsistencies between structs and tightening up and simplifying the tests, I’ve released sgp4 2.7, that fully enables your new `sgp4init()` methods. The tarball is up already, and the wheels are being compiled right now as I type.

Let me know it if starts working on your end once you are able to upgrade! And hopefully contributors will now find the test suite in vastly better shape than when you started work on this new feature. Thanks again.

(One last thing: let me know if I can rephrase any of my above comments to correct anything you found unfriendly; with my head down in the numbers yesterday, I wasn’t sure whether to read your “I'm told that's what astronomers are used to seeing” as merely wry, or as irked or frustrated by something I said. Should I maybe not have cited my vague knowledge of “astronomy convention” in front of a former Flight Director?)

interplanetarychris:
Don't worry about it - attempts at wry humor.

interplanetarychris:
I see it in pypi, everything installs, tests and works fine so far as I can tell at the moment... Thanks!

brandon-rhodes:
Great! Thanks for tying up both loose ends :)

PR #50, opened 2020-02-06
Minor documentation fix

rzimmerman:
The TLE referenced as "Vanguard 1" in the docs looks like it is actually from the ISS.


brandon-rhodes:
Oh, it looks like it is!

PR #55, opened 2020-02-25
Compile wrapper under windows.

mgrutten:
I've made a couple of small changes so that wrapper.cpp compiles under windows (using the compiler that comes with Visual Studio 2019 community edition).

I checked that it still compiles under linux (the version of Ubuntu in the WSL) and that the tests.py run in windows and linux.

### DIFF
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -237,7 +237,7 @@ Satrec_len(PyObject *self) {
 }
 
 static PySequenceMethods SatrecArray_as_sequence = {
-    sq_length : Satrec_len,
+    Satrec_len
 };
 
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thank you! I am going to rework the Windows build this morning so that it fails if it turns out that I am producing wheels with no compiled modules inside — I had not checked whether all the wheels had compiled modules inside or not. Then, I'll merge in this change and see what the result looks like across all platforms (which the standard build doesn't check, to save resources). I'll let you know the result!

brandon-rhodes:
It worked! Thank you. I'll go ahead and try making a new release to bring the benefit of acceleration to Windows folks.

PR #58, opened 2020-04-05
Added jday_datetime() to convert native `datetime` object into JD.

egemenimre:
Added jday_datetime() to parse native `datetime` object. Checks for timezone and converts the time into UTC. May reqire `pytz` and `datetime` added to requirements.txt

### DIFF
--- a/sgp4/functions.py
+++ b/sgp4/functions.py
@@ -5,6 +5,11 @@ modules to offer simple date handling, so this small module holds the
 routines instead.
 
 """
+from datetime import datetime
+
+import pytz
+
+
 def jday(year, mon, day, hr, minute, sec):
      """Return two floats that, when added, produce the specified Julian date.
 
@@ -37,3 +42,44 @@ def jday(year, mon, day, hr, minute, sec):
            + 1721013.5)
      fr = (sec + minute * 60.0 + hr * 3600.0) / 86400.0;
      return jd, fr
+
+def jday_datetime(dtime : datetime):
+    """Return two floats that, when added, produce the specified Julian date.
+
+     The first float returned gives the date, while the second float
+     provides an additional offset for the particular hour, minute, and
+     second of that date.  Because the second float is much smaller in
+     magnitude it can, unlike the first float, be accurate down to very
+     small fractions of a second.
+
+     >>> jd, fr = jday(2020, 2, 11, 13, 57, 0)
+     >>> jd
+     2458890.5
+     >>> fr
+     0.58125
+
+     Note that the first float, which gives the moment of midnight that
+     commences the given calendar date, always carries the fraction
+     ``.5`` because Julian dates begin and end at noon.  This made
+     Julian dates more convenient for astronomers in Europe, by making
+     the whole night belong to a single Julian date.
+
+    The input is a native `datetime` object. Timezone  of the input is converted internally to UTC.
+
+     This function is a simple translation to Python of the C++ routine
+     ``jday()`` in Vallado's ``SGP4.cpp``.
+
+    """
+    # Convert to UTC time
+    dtime_utc = dtime.astimezone(pytz.UTC)
+    year = dtime_utc.year
+    mon = dtime_utc.month
+    day = dtime_utc.day
+
+    hr = dtime_utc.hour
+    minute = dtime_utc.minute
+    sec = dtime_utc.second + dtime_utc.microsecond / 1e6
+
+    return jday(year, mon, day, hr, minute, sec)
+
+
--- a/sgp4/functions.py
+++ b/sgp4/functions.py
@@ -5,9 +5,7 @@ modules to offer simple date handling, so this small module holds the
 routines instead.
 
 """
-from datetime import datetime
-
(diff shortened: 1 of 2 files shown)


egemenimre:
A sample run could be:

`from datetime import datetime, tzinfo

import pytz

from sgp4.functions import jday_datetime

if __name__ == '__main__':

    datetime_now: datetime = datetime.now(tz=pytz.timezone("Turkey"))
    print(f"Time in Turkey: {datetime_now} {datetime_now.tzinfo}")

    datetime_now_utc = datetime_now.astimezone(pytz.UTC)
    print(f"Time in UTC   : {datetime_now_utc} {datetime_now_utc.tzinfo}")

    jd, fr = jday_datetime(datetime_now_utc)

    print(f"Julian Date part: {jd}")
    print(f"Fraction part   : {fr}")`

sgp4/functions.py
```diff
@@ -5,6 +5,11 @@
 routines instead.
 
 """
+from datetime import datetime
+
+import pytz
```
brandon-rhodes, line 10: Most Python users, alas, don't have `pytz` installed, so this will raise an error. And I would prefer a library like this that's so simple to not require time zone importation, when most users don't need that.

Try this instead, it's how I solved the same problem in Skyfield:

    class UTC(tzinfo):
        'UTC'
        zero = timedelta(0)
        def utcoffset(self, dt):
            return self.zero
        def tzname(self, dt):
            return 'UTC'
        def dst(self, dt):
            return self.zero

    utc = UTC()

sgp4/functions.py
```diff
@@ -37,3 +42,44 @@ def jday(year, mon, day, hr, minute, sec):
...
+     >>> fr
+     0.58125
+
+     Note that the first float, which gives the moment of midnight that
+     commences the given calendar date, always carries the fraction
+     ``.5`` because Julian dates begin and end at noon.  This made
+     Julian dates more convenient for astronomers in Europe, by making
+     the whole night belong to a single Julian date.
+
+    The input is a native `datetime` object. Timezone  of the input is converted internally to UTC.
```
brandon-rhodes, line 67: Could you clean up the whitespace a little, and wrap it to the line length of the paragraphs around it? Thanks!

sgp4/functions.py
```diff
@@ -37,3 +42,44 @@ def jday(year, mon, day, hr, minute, sec):
...
+    """
+    # Convert to UTC time
+    dtime_utc = dtime.astimezone(pytz.UTC)
+    year = dtime_utc.year
+    mon = dtime_utc.month
+    day = dtime_utc.day
+
+    hr = dtime_utc.hour
+    minute = dtime_utc.minute
+    sec = dtime_utc.second + dtime_utc.microsecond / 1e6
```
brandon-rhodes, line 81: Would `* 1e-6` be slightly more elegant? I might be overly influenced by the old days, when floating point multiply was several times faster than division :)

sgp4/functions.py
```diff
@@ -37,3 +42,44 @@ def jday(year, mon, day, hr, minute, sec):
...
+
+     Note that the first float, which gives the moment of midnight that
+     commences the given calendar date, always carries the fraction
+     ``.5`` because Julian dates begin and end at noon.  This made
+     Julian dates more convenient for astronomers in Europe, by making
+     the whole night belong to a single Julian date.
+
+    The input is a native `datetime` object. Timezone  of the input is converted internally to UTC.
+
+     This function is a simple translation to Python of the C++ routine
```
brandon-rhodes, line 69: You can remove this statement as it's not actually true of this new routine. :)

sgp4/functions.py
egemenimre: OK, fair point. I found this solution that uses the native `datetime`. I will implement it without pytz but with the same local time-to-UTC functionality.

```
from datetime import datetime, timezone

    datetime_now: datetime = datetime.now()
    print(f"Time in Turkey: {datetime_now} {datetime_now.tzinfo}")

    datetime_now_utc = datetime_now.astimezone(timezone.utc)
    print(f"Time in UTC   : {datetime_now_utc} {datetime_now_utc.tzinfo}")
```

egemenimre:
All changes made, though there is a conflict in imports. Hmm...

brandon-rhodes:
Since the whole test suite was reformatted, I merged your test into the code myself, so you're now landed!

I also added Python 2 compatibility. Since the CI reporting was down (mysteriously; it turns out I needed to cancel the Travis CI integration then re-authorize Travis CI to use GitHub before build results started flowing again), I went ahead and did that myself.

brandon-rhodes:
I'm undecided about where the routine should live. Its current home in `functions.py` means it gets pulled in to every import of `sgp4`, but for folks using higher-level libraries like Skyfield that have their own date processing, this routine and its timezone are redundant. Which is also true for folks who don't use datetimes.

When I write up the docs for it before the next release, I'll have to decide. I might put it in a little `conveniences.py` module if at that time I still feel a tug in the direction of keeping the core packages simple.

PR #59, opened 2020-04-06
Added export TLE functionality to `Satrec`

egemenimre:
Adds "export TLE" functionality to `Satrec`. This enables modification of the TLE and safely exporting it.

See sample code below.

```
from sgp4.model import Satrec

if __name__ == '__main__':
    line1 = "1 00005U 58002B   00179.78495062  .00000023  00000-0  28098-4 0  4753"
    line2 = "2 00005  34.2682 348.7242 1859667 331.7664  19.3264 10.82419157413667"

    # line1 = "1 16925U 86065D   06151.67415771  .02550794 -30915-6  18784-3 0  4486"
    # line2 = "2 16925  62.0906 295.0239 5596327 245.1593  47.9690  4.88511875148616"

    print(line1)
    print(line2)

    satrec = Satrec.twoline2rv(line1, line2)

    new_line1, new_line2 = satrec.export_tle()

    print(new_line1)
    print(new_line2)
```

### DIFF
--- a/sgp4/export.py
+++ b/sgp4/export.py
@@ -1,0 +1,101 @@
+"""
+Export orbit data to a Two-Line-Element representation.
+"""
+from sgp4.io import compute_checksum
+
+from sgp4.model import Satrec
+
+
+def export_tle(satrec):
+    """
+     Generate the TLE for the data in the `Satrec` object.
+
+    Forms the two string lines of the TLE and returns them as a Tuple.
+
+    Parameters
+    ----------
+    satrec : Satrec
+        Object that holds the orbit information
+
+    Returns
+    -------
+    (line1, line2) : (str, str)
+    """
+
+    # Define constants
+    from math import pi
+    deg2rad = pi / 180.0  # 0.0174532925199433
+    xpdotp = 1440.0 / (2.0 * pi)  # 229.1831180523293
+
+    # --------------------- Start generating line 1 ---------------------
+
+    # Build the list by appending successive items
+    line_as_list = ["1 "]
+
+    # Pad the `satnum` entry with zeros
+    if len(str(satrec.satnum)) <= 5:
+        line_as_list.append(str(satrec.satnum).zfill(5))
+
+    # Add classification code (use "U" if empty)
+    line_as_list.append((satrec.classification.strip() or "U") + " ")
+
+    # Add int'l designator and pad to 8 chars
+    line_as_list.append(satrec.intldesg.ljust(8, " ") + " ")
+
+    # Add epoch year and days in YYDDD.DDDDDDDD format
+    line_as_list.append(str(satrec.epochyr).zfill(2) + "{:012.8f}".format(satrec.epochdays) + " ")
+
+    # Add First Time Derivative of the Mean Motion (don't use "+")
+    line_as_list.append("{0: 8.8f}".format(satrec.ndot * (xpdotp * 1440.0)).replace("0", "", 1) + " ")
+
+    # Add Second Time Derivative of Mean Motion (don't use "+")
+    # Multiplication with 10 is a hack to get the exponent right
+    line_as_list.append("{0: 4.4e}".format((satrec.nddot * (xpdotp * 1440.0 * 1440)) * 10).replace(".", "")
+                        .replace("e+00", "-0").replace("e-0", "-") + " ")
+
+    # Add BSTAR
+    # Multiplication with 10 is a hack to get the exponent right
+    line_as_list.append("{0: 4.4e}".format(satrec.bstar * 10).replace(".", "").replace("e+00", "+0").replace("e-0", "-") + " ")
+
+    # Add Ephemeris Type and Element Number
+    line_as_list.append("{} ".format(satrec.ephtype) + str(satrec.elnum).rjust(4, " "))
+
+    # Join all the parts and add the Checksum
+    line1 = ''.join(line_as_list)
+    line1 += str(compute_checksum(line1))
+
+    # --------------------- Start generating line 2 ---------------------
+
+    # Reset the str array
+    line_as_list = ["2 "]
+
+    # Pad the `satnum` entry with zeros
+    if len(str(satrec.satnum)) <= 5:
+        line_as_list.append(str(satrec.satnum).zfill(5) + " ")
+
+    # Add the inclination (deg)
+    line_as_list.append("{0:8.4f}".format(satrec.inclo / deg2rad).rjust(8, " ") + " ")
+
+    # Add the RAAN (deg)
+    line_as_list.append("{0:8.4f}".format(satrec.nodeo / deg2rad).rjust(8, " ") + " ")
+
+    # Add the eccentricity (delete the leading zero an decimal point)
+    line_as_list.append("{0:8.7f}".format(satrec.ecco).replace("0.", "") + " ")
+
+    # Add the Argument of Perigee (deg)
+    line_as_list.append("{0:8.4f}".format(satrec.argpo / deg2rad).rjust(8, " ") + " ")
+
+    # Add the Mean Anomaly (deg)
+    line_as_list.append("{0:8.4f}".format(satrec.mo / deg2rad).rjust(8, " ") + " ")
+
+    # Add the Mean Motion (revs/day)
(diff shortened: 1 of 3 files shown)


ckuethe:
Looks like the use of f-strings makes this diff incompatible with python2.7. Yeah, I know that's deprecated but other PRs are still maintaining compatibility (https://github.com/brandon-rhodes/python-sgp4/pull/49#pullrequestreview-353561106)

I've been building my TLEs using `str.format()`...

egemenimre:
Argh. I'm fairly new to Python, so I'm rather unfamiliar with this Python 2.X business. I'm open to suggestions - I can rewrite it in the str.format() and recommit. Is it 3.X and 2.X compatible without deprecation issues?

I really think modifying and exporting TLEs would be a good addition for a lot of the research folks out there.

brandon-rhodes:
Yes, switching to str.format() should be compatible with all the versions of Python we support.

sgp4/model.py
```diff
@@ -89,6 +89,90 @@ def sgp4_array(self, jd, fr):
...
+        xpdotp = 1440.0 / (2.0 * pi)  # 229.1831180523293
+
+        # --------------------- Start generating line 1 ---------------------
+
+        # Strings are immutable, so rather than replacing chars, we will build it
+        line1 = "1 "
+
+        # Pad the `satnum` entry with zeros
+        if len(str(self.satnum)) <= 5:
+            line1 += str(self.satnum).zfill(5)
```
brandon-rhodes, line 113: Building a longer and longer line tends to be inefficient. Instead, try creating a list `strings = []` and append to it all these little pieces of string, and then you can `''.join(strings)` at the end to all at once build a big string out of them.

sgp4/model.py
```diff
@@ -89,6 +89,90 @@ def sgp4_array(self, jd, fr):
         r.shape = v.shape = len(jd), 3
         return e, r, v
 
+    def export_tle(self):
```
brandon-rhodes, line 92: Since this method does not access any private attributes of the TLE, or modify its state, I think we can simplify here by making it a plain function. Could you create a new module `export.py` and place this there as a plain function `export_tle(satrec)` that takes the satrec as its argument? Thanks!

egemenimre: Done! I will commit this file along with the str.format() changes you suggested.

sgp4/model.py
egemenimre: Thanks for the suggestion. I have now (locally) modified the string building methodology to this `append()` method. Will do the `str.format()` thing and recommit.

egemenimre:
OK, so I have implemented the changes:
- Used Python 2.X friendly `str.format`
- Opened a new `export.py` file
- Added a test that reads the SGP4-VER.TLE file and runs a round-trip test (two-lines to `Satrec` to two-lines again). There are a few TLEs that fail, some due to BSTAR sign inconsistencies (though the resulting `Satrec` object is the same) and others due to checksum errors already marked in the file.

sgp4/export.py
```diff
@@ -0,0 +1,101 @@
...
+    satrec : Satrec
+        Object that holds the orbit information
+
+    Returns
+    -------
+    (line1, line2) : (str, str)
+    """
+
+    # Define constants
+    from math import pi
```
brandon-rhodes, line 26: Typically in Python imports are moved to the top-level, so the import statement doesn't have to run over again every time the function is called. Try moving it out to the module level, and try reading PEP-8 which has good ideas about how to format imports. Thanks!

sgp4/export.py
```diff
@@ -0,0 +1,101 @@
...
+    deg2rad = pi / 180.0  # 0.0174532925199433
```
brandon-rhodes, line 27: Instead of computing this every time the function is called, try moving it outside the function as a constant `_deg2rad` out at module level. Try the same thing with the constant on the next line, too.

sgp4/export.py
```diff
@@ -0,0 +1,101 @@
...
+    xpdotp = 1440.0 / (2.0 * pi)  # 229.1831180523293
+
+    # --------------------- Start generating line 1 ---------------------
+
+    # Build the list by appending successive items
+    line_as_list = ["1 "]
+
```
brandon-rhodes, line 34: Go ahead right here and say:
```
append = line_as_list.append
```

Then all the subsequent calls can just say `append()` instead of repeating the name `line_as_list`. Oh, and, if you'd like, feel free to call it just `strings` or `pieces`. Readers typically will even with a simple single-word name recognize the maneuver and know what's going on. (And you can do the same down where you work on line 2.)

sgp4/tests.py
```diff
@@ -31,6 +31,53 @@
...
+            # print("satrec: " + out_line2)
+
+            # Skip these lines, known errors
+            # Resulting TLEs are equivalent (same Satrec object) but they are not the same
+            # 25954: BSTAR = 0 results in a negative exp, not positive
+            # 29141: BSTAR = 0.13519 results in a negative exp, not positive
+            # 33333: Checksum error as expected on both lines
+            # 33334: Checksum error as expected on line 1
+            # 33335: Checksum error as expected on line 1
+            expected_errs_line1 = [25954, 29141, 33333, 33334, 33335]
```
brandon-rhodes, line 73: It will take a little time to rebuild each of these lists every time through this loop. So:

1. Move the definitions up outside the loop, above the `for`. That way, the lists only get created once.
2. The `in` operation on a list needs to loop through the whole sequence. Try using sets here instead. Thanks!

sgp4/tests.py
```diff
@@ -31,6 +31,53 @@
...
+            expected_errs_line2 = [33333, 33335]
+
+            if satrec.satnum not in expected_errs_line1:
```
brandon-rhodes, line 76: Nice, I like how cleanly you guarded against the "bad" test values here, and how you test the good line even in cases where you know the other line is bad!

sgp4/tests.py
```diff
@@ -31,6 +31,53 @@
...
+
+            # print("file :  " + line1)
+            # print("satrec: " + out_line1)
+            #
+            # print("file :  " + line2)
+            # print("satrec: " + out_line2)
+
+            # Skip these lines, known errors
+            # Resulting TLEs are equivalent (same Satrec object) but they are not the same
+            # 25954: BSTAR = 0 results in a negative exp, not positive
```
brandon-rhodes, line 68: Thanks for the detailed comments!

sgp4/export.py
egemenimre: Argh, I must have chosen the "import locally" option in PyCharm by mistake. To be fixed right away.

sgp4/export.py
egemenimre: I feel like I'm getting a one-to-one training from a Python expert, cheers for that. I know my way around Java, but I'm new to all things Pythonic. :)

sgp4/tests.py
egemenimre: 1. That was me being silly, fixed.
2. Changed the following to `set`:
```
        expected_errs_line1 = set([25954, 29141, 33333, 33334, 33335])
        expected_errs_line2 = set([33333, 33335])
```

brandon-rhodes:
The changes are looking good! Remember to "request review" when you're done changing things so I know to come back — otherwise I might jump in and start re-reviewing when more changes are in fact coming.

egemenimre:
Thanks for the useful tips, the code is updated for your (hopefully final) review.

brandon-rhodes:
Thanks again for the contribution! I'll go add some docs.

egemenimre:
Thanks! I forgot a silly `print` line in tests.py and now I have pushed it to the branch.

brandon-rhodes:
Let me know if you want credit for that module, I'd be happy to add your name and a link to your organization at the top of the file.

egemenimre:
Just add my name and a link to my github page - it would be much appreciated. And thank you for the useful sgp4 module!

brandon-rhodes:
Sure! What's your full name — or should I just use your GitHub username?

egemenimre:
The name is Egemen Imre. :)

brandon-rhodes:
Done!

Issue #65, opened 2020-06-25
Support OMM file format


benjaminglass1:
As documented at [Celestrak's website](https://celestrak.com/NORAD/documentation/gp-data-formats.php), an XML/JSON based format called OMM (spec: https://public.ccsds.org/Pubs/502x0b2c1.pdf) is being pushed as a more modern replacement for the TLE format. Since the parameters are still reported in terms of sgp4 elements, and given the dominance of CelesTrak as a source of orbit information, it seems reasonable imo to add support to this project for the OMM format.

I'd be happy to open a PR for this with some guidance on how you'd like this structured. Would it be more reasonable to:

1. Add this functionality into `io.py`?
2. Create a new submodule, `omm.py` or something similar?
3. Something else?

This is my first time trying to contribute to an open-source project, so apologies if there's something I'm missing.

ckuethe:
Go ahead and just start coding something and make a PR. You'll probably get more useful feedback with actual code that we can talk about. If it was me, I'd start by prototyping it with just the JSON format because I have an unreasonable loathing of XML, and because it's way easier to ingest JSON objects into python than XML. I kind of accidentally independently invented a crappy version of this because I wanted to store TLEs in a MongoDB instance and breaking out all the parameters into fields in a document made it very easy to work with. So thanks for highlighting that there is a standard that I can develop against. 

Part of me really wants to say "XML should f*ck off and die in a hard drive shredder". I was recently developing some monitoring for my home photovoltaic system that has an API that is all JSON... except for one stupid XML endpoint, and there's no good reason for it. The code to handle that one endpoint is way more complicated than all the rest of the endpoint handlers. `xmltodict` can be one useful tool, if my rant hasn't discouraged you from fully implementing OMM including XML.

Parsing the JSON version is literally a one-liner, and once you know the OMM JSON reader works you can extend it to the other formats and cross-validate. It looks like `Satrec.twoline2rv(line1, line2)` is the way to create a satellite from a TLE, so maybe `Satrec.omm(format=Satrec.JSON, data=buf)` might be a reasonable interface. Or maybe `ommjson`, `ommcsv`,  `ommxml`, 
 so that it it absolutely clear to users what the input format is supposed to be. 

One thing to consider is what happens when I do something like the example below and feed in a JSON object that contains multiple sets of parameters... food for thought :)

```
>>> import requests
>>> from pprint import pprint
[...]

PR #66, opened 2020-07-01
Test for similar projected state between TLE and XML - Issue #65

glangford:
Test/Evaluation code - for Stations in stations.txt and the XML equivalent data published by Celestrak, create 'twin' EarthSatellite objects using each source and verify that they have a similar position at a set time. For the included TLE and XML files, the positions of all Stations are essentially identical. (This pull request originated at python-skyfield and is being moved here).


brandon-rhodes:
Thank you for moving this over, and for putting together working code that from raw XML produces working satellite objects!

I plan on making several changes to the code for efficiency once it’s merged, but I don’t think it would be a good use of your time for me to force you to transform the code — you’ve provided a working prototype with a test to keep it working during any refactoring, which is a great starting point that can be merged.

If you could perform just a couple of quick pivots:

1. Could you reduce the satellite files to only 3 satellites or less? If even a single satellite is read in correctly, my guess is that all satellites will work correctly, and that it will only burn time for folks waiting for the test suite to test a whole file full.
2. Could you rename the module? Python cannot import files with a dash in the name, so your routine cannot in its present state be used by any users of `sgp4` in Python. Renaming `tle-xml.py` to something like `omm_format.py` should do the trick, and will also advertise the more official name of the format ("Orbit Mean-Elements Message", it looks like, is the expansion of the acronym).
3. As a bonus you could move the test into the `sgp4/tests.py` file, but I can merge either way and move that logic over myself.

Thanks!

glangford:
To test the whole satellites file takes ~ 1s on my machine so I opted for completeness (given that fields and contents can vary) - but I can certainly trim it down to 3. I will also rename!

brandon-rhodes:
> To test the whole satellites file takes ~ 1s on my machine…

The entire test suite for `sgp4` currently takes around 0.4s on my machine, so an additional second would be a large regression. When developing, I typically re-run the test suite several times per minute as I refactor code and hit "Save", so keeping the suite tight helps to keep development productive. Thanks!

glangford:
I'm not sure how you would like the code within `sgp4/tests.py` so I left it separate, and it is ready for merge.

brandon-rhodes:
Thanks for making those adjustments! Hopefully I'll have time to turn this into a fully documented feature within the next week or so.

Issue #70, opened 2020-11-07
Out of range produce invalid TLEs


astrojuanlu:
The TLE FAQ specifies that:

> Fields 2.3, 2.4, 2.6, and 2.7 all have units of degrees and can range from 0 up to 360 degrees—field 2.3 (inclination) only goes up to 180 degrees.

However, if one creates a `Satrec` object with, say, `argpo=-1.90861`:

- `sgp4init` does not complain
- `rv2twoline` produces an invalid TLE!

For example:

```
1 99510U 20999B   20311.53996093  .00007856  00000-0  10000-4 0  9995  
2 99510  97.2675  23.7519 0007535 -109.3554278.9614 15.31068184    67 
```

Should python-sgp4 perform any sort of validation?

astrojuanlu:
(Edit: typo, `argpo` is specified in radians, but I copy-pasted the value in degrees)

brandon-rhodes:
Good question. Since the library is supposed to be a thin wrapper around the real C++ library, I am not sure it would be in the spirit of the library for `sgp4init()` to perform any checks that the underlying `sgp4init()` routine does not itself perform.

But that logic would not apply to any routines that aren't members of the underlying library — they could change without raising any surprises, since there is no C++ version of them. But first we should get clear on what routine we are talking about; I don't see a `rv2twoline()` function anywhere? Do you mean `export_tle()`?

astrojuanlu:
Oh, I realized `rv2twoline` is only present in this fork by @enritoomey https://github.com/enritoomey/python-sgp4/blob/95b08fdfc8131bf6b7c81cd1211ba3eaa032349a/sgp4/io.py#L254 (:wave:). But yes, `export_tle` introduced in #59 (`sgp4>=2.6`) has this same problem:

```
>>> lines = """1 26975U 78066F   06174.85818871  .00000620  00000-0  10000-3 0  6809
... 2 26975  68.4714 236.1303 5602877 123.7484 302.5767  2.05657553 67521"""
>>> from sgp4.exporter import export_tle
>>> from sgp4.earth_gravity import wgs84
>>> from sgp4.io import twoline2rv
>>> sat = twoline2rv(*lines.splitlines(), wgs84)
>>> export_tle(sat)
('1 26975U 78066F   2006174.85818871  .00000620  00000-0  10000-3 0  6803', '2 26975  68.4714 236.1303 5602877 123.7484 302.5767  2.05657553 67521')
>>> sat.argp
[...]

PR #72, opened 2020-11-14
Raise error in export_tle if angles are invalid

astrojuanlu:
Fix gh-70.

### DIFF
--- a/sgp4/exporter.py
+++ b/sgp4/exporter.py
@@ -63,12 +63,12 @@ def export_tle(satrec):
 
     # Add the inclination (deg)
     if not 0 <= satrec.inclo <= pi:
-        raise ValueError("Inclination must be between 0 and pi, got %f", satrec.inclo)
+        raise ValueError("Inclination must be between 0 and pi, got %r", satrec.inclo)
     append("{0:8.4f}".format(satrec.inclo / _deg2rad).rjust(8, " ") + " ")
 
     # Add the RAAN (deg)
     if not 0 <= satrec.nodeo <= 2 * pi:
-        raise ValueError("RAAN must be between 0 and 2 pi, got %f", satrec.nodeo)
+        raise ValueError("RAAN must be between 0 and 2 pi, got %r", satrec.nodeo)
     append("{0:8.4f}".format(satrec.nodeo / _deg2rad).rjust(8, " ") + " ")
 
     # Add the eccentricity (delete the leading zero an decimal point)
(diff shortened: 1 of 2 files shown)


sgp4/exporter.py
```diff
@@ -62,18 +62,26 @@ def export_tle(satrec):
         append(str(satrec.satnum).zfill(5) + " ")
 
     # Add the inclination (deg)
+    if not 0 <= satrec.inclo <= pi:
+        raise ValueError("Inclination must be between 0 and pi, got %f", satrec.inclo)
```
brandon-rhodes, line 66: Let’s try `%r` instead of `%f` for these value checks, in case anyone is ever using the old Python satrec and inserts a value that’s not a float.

sgp4/tests.py
```diff
@@ -207,6 +207,19 @@ def test_tle_export():
         if satrec.satnum not in expected_errs_line2:
             assertEqual(out_line2, line2)
 
+def test_export_tle_raises_error_for_out_of_range_angles():
+    # See https://github.com/brandon-rhodes/python-sgp4/issues/70
+    sat = Satrec()
+
+    for angle in 'inclo', 'nodeo', 'argpo', 'mo':
+        WRONG_VANGUARD_ATTRS = VANGUARD_ATTRS.copy()
```
brandon-rhodes, line 215: For a local variable, try flipping to lowercase — so: `wrong_vanguard_attrs` for the value here.

sgp4/tests.py
```diff
@@ -207,6 +207,19 @@ def test_tle_export():
         if satrec.satnum not in expected_errs_line2:
             assertEqual(out_line2, line2)
 
+def test_export_tle_raises_error_for_out_of_range_angles():
+    # See https://github.com/brandon-rhodes/python-sgp4/issues/70
+    sat = Satrec()
```
brandon-rhodes, line 212: Go ahead and move the creation of `sat` inside the loop, so that folks reading the test later can be assured that there's no state being invisibly carried from one iteration of the loop to the next.

brandon-rhodes (changes requested):
Thank you! This looks good, and I'm about ready to merge. There are three quick comments but once those are addressed you'll be all done!

sgp4/exporter.py
astrojuanlu: Done!

sgp4/tests.py
astrojuanlu: Good idea, done!

astrojuanlu:
Thanks for your prompt review! I addressed your comments.

brandon-rhodes:
Thanks for the contribution!

PR #79, opened 2020-12-18
Add export_omm function

astrojuanlu:
See #65.

No tests yet.

### DIFF
--- a/sgp4/exporter.py
+++ b/sgp4/exporter.py
@@ -5,6 +5,7 @@ Contributed by Egemen Imre https://github.com/egemenimre
 """
 from math import pi
 from sgp4.io import compute_checksum
+from sgp4.conveniences import sat_epoch_datetime
 
 # Define constants
 _deg2rad = pi / 180.0  # 0.0174532925199433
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
I should follow up on this, since it's a new year: does the “No tests yet.” comment mean I shouldn't look over this yet, because tests are forthcoming and it's a work in progress? I think that was my assumption when I saw that comment, but maybe you are waiting on me instead. Let me know either way, thanks!

brandon-rhodes:
I'm releasing a new version this morning, so I decided to go ahead and accept this PR (so, thank you!) and write the test myself. The only error I found was an extra hyphen in `OBJECT_ID`, which I'll go ahead and remove.

astrojuanlu:
Hi @brandon-rhodes! Sorry for the delay - to be honest, my "no tests yet" message was an ambiguous mix of "I will continue working on this soon" (other stuff sucked my attention these past weeks) and "have a look in case I am doing something very stupid or the code does not belong here".

If you still want, I think I can add some tests these weekend. But don't let that stop you from releasing!

astrojuanlu:
Also, thanks for merging it! 😊

PR #83, opened 2021-01-18
Add a note about SGP4 ignoring ndot and nddot

pfandzelter:
If you skip past parts of the package documentation and on to the _Providing your own elements_ section you'll have trouble understanding some of the parameters, particularly the `ndot` and `nddot` ones, which, according to other parts of the documentation, are ignored by SGP4. To make getting started with this API easier for newbies (i.e., me), I've added a small note that these two parameters are ignored by SGP4.
I've also added named parameters which makes it easier to use the example code in projects.

### DIFF
--- a/sgp4/__init__.py
+++ b/sgp4/__init__.py
@@ -352,8 +352,8 @@ satellite object to reset it to those new elements.
 ...     5,               # satnum: Satellite number
 ...     18441.785,       # epoch: days since 1949 December 31 00:00 UT
 ...     2.8098e-05,      # bstar: drag coefficient (/earth radii)
-...     6.969196665e-13, # ndot: ballistic coefficient (revs/day)
-...     0.0,             # nddot: second derivative of mean motion (revs/day^3)
+...     6.969196665e-13, # ndot: ballistic coefficient (ignored)
+...     0.0,             # nddot: second derivative of mean motion (ignored)
 ...     0.1859667,       # ecco: eccentricity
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
1. Alas, the accelerated version of that routine does not accept keyword arguments; you probably tested against the unaccelerated fallback Python code. Here are your branch test results: https://travis-ci.org/github/brandon-rhodes/python-sgp4/builds/754995964
2. At some point the old Travis CI integration, which is clearly now broken (the results used to be posted back to pull requests for authors), will need to be replaced with GitHub Actions. Until then you will have to look for build results on Travis manually, or run `tox` locally. I'll go add an issue so that potential contributors know that CI support is currently missing.
3. Let’s try keeping the lines to less than 80 characters, for folks with standard terminal windows. But I think at least the word `(ignored)` would fit next to those two parameters, so the same information is conveyed in the comments as in the more full descriptions above?

pfandzelter:
Thanks for the quick review @brandon-rhodes. I did in fact get an email from Travis later that informed me about my breaking changes.
I edited my changes down to just the `(ignored)` part, which should help it pass the tests and work on standard terminals.

At least I found out that I'm not using the high performance version today!

brandon-rhodes:
Oh, I didn't know Travis was at least still in the email business!

We'll see if this adjustment helps users understand those numbers better. Note that the online docs won't update until the next release, because PyPI now disallows edits or fixes to the descriptions of released packages.

Issue #84, opened 2021-01-18
Move to GitHub actions


brandon-rhodes:
Potential contributors need feedback from CI test builds, but with Travis CI at its end-of-life, build results are no longer being posted back to Pull Requests.

1. Developers can run `tox` locally, if they `pip install tox` first, to at least see whether the tests pass locally.
2. And longer-term, a volunteer will need to replace the project `.travis.yml` with the GitHub Actions equivalent to make CI results available again.

Anyone interested in tackling the project can comment here on this issue expressing their interest. In the meantime, contributors should use `tox` locally.

mworion:
@brandon-rhodes 
If you don't mind, I'll give it a try during the next 1 or two weeks. I never did use travis, but I do use GitHub actions in my project. So I will find out. My guess is that you would like to run the tests each commit and each PR.
Michel

brandon-rhodes:
@mworion — Thank you for the offer! I'll assign you this issue officially. Yes, commits and PRs. Like the `tox` setup and the current Travis CI setup, the project will need both at least a Python 2.7 run and a Python 3.something run where the latter makes sure that the project compiles under Python 3 instead of falling back to the Python version. (See `setup.py` which might detect when it's on Travis?)

I'll watch the issue over the next several weeks in case you have questions.

mworion:
@brandon-rhodes:
linting with pyflakes is running on all demanded  python versions
sgp4/tests are running on all demanded  python versions
testing build package done as well

some questions:
- do I have to build and test on ppc64le as well? (difficult)
- need to test on different os (windows, Ubuntu, macOS)? (easy)
- what about having the the speedup version. How could I see if the c code is running ? (no clue so far)
- Building wheels afterwards (or separate action) (duable)
- Upload possibility to PyPi ? (easy step, your have to configure)

brandon-rhodes:
> do I have to build and test on ppc64le as well? (difficult)

Don't spend time on it, no. If there are third parties for whom that build is crucial, they can run as part of their packaging suite, or maybe will be interested in contributing the work towards having that platform in GitHub Actions.

> need to test on different os (windows, Ubuntu, macOS)? (easy)

Sure, let's go for it. It's more compute on each commit, but development is pretty much done at this point, so I'm not anticipating high volume — and if folks are going to break things under other OS's, it lets them discover that rather than my not learning about it til release time :)

> what about having the the speedup version. How could I see if the c code is running ? (no clue so far)

[...]

Issue #69, opened 2020-10-16
precision tests too tight for "exotic" platforms


bnavigator:
Cf. https://github.com/skyfielders/python-skyfield/issues/411

- i586
```
[   37s] ________________ test_all_three_gravity_models_with_twoline2rv _________________

[   37s] >       assert_wgs72old(Satrec.twoline2rv(LINE1, LINE2, WGS72OLD))
[   37s] E       AssertionError: -3754.2514732427567 != -3754.251473242793 within 12 places (3.637978807091713e-11 difference)


[   37s] _________________ test_all_three_gravity_models_with_sgp4init __________________
[   37s]         sat.sgp4init(WGS72OLD, 'i', VANGUARD_ATTRS['satnum'], VANGUARD_EPOCH, *args)
[   37s] >       assert_wgs72old(sat)
[   37s] E       AssertionError: -3754.2514732427567 != -3754.251473242793 within 12 places (3.637978807091713e-11 difference)
```

- aarch64, ppc64, ppc64le, s390x all produce the same numbers for WGS84
```
[   50s] >       assert_wgs84(Satrec.twoline2rv(LINE1, LINE2, WGS84))
[   50s] E       AssertionError: 4719.227897029575 != 4719.227897029576 within 12 places (9.094947017729282e-13 difference)

[   50s]         sat.sgp4init(WGS84, 'i', VANGUARD_ATTRS['satnum'], VANGUARD_EPOCH, *args)
[   50s] >       assert_wgs84(sat)
[   50s] E       AssertionError: 4719.227897029575 != 4719.227897029576 within 12 places (9.094947017729282e-13 difference)
```

brandon-rhodes:
I have just landed an adjustment of the gravity parameter test precision for those platforms — let me know if it looks like it will solve your use cases.

brandon-rhodes:
(I tried adding a 32-bit test container in the hopes of simulating your i586 results, the kind of Docker container I use for the same thing in Skyfield, but the tests passed just fine, so I'm not able to reproduce the problem locally.)

bnavigator:
I applied your patch and the previously reported tests pass.

However, the following **new** failures emerged (on i586 only):

```
[   26s] =================================== FAILURES ===================================
[   26s] ________________ test_satrec_against_tcppver_using_julian_dates ________________
[   26s] >       run_satellite_against_tcppver(Satrec.twoline2rv, invoke, [1,1,6,6,4,3,6])

[...]

PR #85, opened 2021-01-21
GitHub Actions for testing, building wheels and upload to PyPI. Addressing #69 and #84.

mworion:
@brandon-rhodes:
PR, please review. Compiler flag is in, test for windows specific carriage return is in. Test run in my environment OK. I tested PyPI upload to test.pypi.org, went also OK. Wheels are build (except aarch64) like it is actually in travis.
GitHub does not support python2.6 nor does cibuilwheels support builds <python3.5. So I left it like this.
Configuration for PyPI upload has to be done (secret).
Hope this helps.
Michel

### DIFF
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -70,13 +70,13 @@ jobs:
       # installation before running the tests
       
       - name: install_deps
-        run: python -m pip install pyflakes
+        run: python -m pip install numpy twine pyflakes
       
       - name: build_sdist
         run: python setup.py sdist
       
-      #- name: check_metadata
-      #  run: twine check dist/*
+      - name: check_metadata
+        run: twine check dist/*
 
       - name: upload_sdist
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9'
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -38,18 +38,19 @@ jobs:
   # This action is split into three jobs:
   # - Building the distribution linting and testing without acceleration
   # - Building the wheels for the distribution and testing with acceleration
-  # - Uploading the artifacts to PyPI package
-  # Each job build on top of the o
+  # - Uploading the artifacts to PyPI package if branch is release
   # The uploading job needs all tests to be finished without error.
   
   build_test_dist:
     # Build the distribution in a matrix. Jobs are done in parallel.
-    # The formal distro which is uploaded will be build on ubuntu-latest for 
-    # python 3.9. which is the recent one.
+    # The formal distro which is uploaded to PyPI will be build on
+    # ubuntu-latest for python 3.9, which is the recent one.
+    # As in dist build no compilation takes place, we run all tests
+    # in not accelerated mode.
     
     runs-on: ${{ matrix.os }}
     strategy:
-      max-parallel: 20
+      max-parallel: 18
       matrix:
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-latest, windows-latest, macos-latest]
@@ -105,6 +106,8 @@ jobs:
     # Reference: https://cibuildwheel.readthedocs.io/en/stable/
     # OS: Windows, Linux and macOS
     # Python: versions 3.4 - 3.9
+    # As all build wheels are installed after build, the tests run in
+    # accelerated mode only.
 
     runs-on: ${{ matrix.os }}
     needs: [build_test_dist]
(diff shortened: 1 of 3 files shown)


.github/workflows/test_package.yml
```diff
@@ -0,0 +1,191 @@
...
+      - name: copy_file
+        run: mv dist/sgp4*.* dist/sgp4.tar.gz
+
+      - name: install_dist
+        run: |
+          pip install dist/sgp4.tar.gz
+          python -c "from sgp4.api import accelerated; print(accelerated)"
+
+      - name: run_tests
+        run: python -m sgp4.tests
```
brandon-rhodes, line 100: Could you also add the `python -c` to `print(accelerated)` here as well? I have had the problem before of `python -m` choosing to import `sgp4` from the current directory, instead of using the compiled version, and `print(accelerated)` helps me determine if I'm really using the compiled version. (Or, does the `run_tests` step happen in a different directory, not the repository?)

sgp4/tests.py
```diff
@@ -536,6 +536,7 @@ def run_satellite_against_tcppver(twoline2rv, invoke, expected_errors):
     # output in tcppver.out.
 
     data = get_data(__name__, 'tcppver.out')
+    data = data.replace(b'\r', b'')
```
brandon-rhodes, line 539: Nice, I'm glad this fix wound up being simple!

setup.py
```diff
@@ -29,7 +29,7 @@
         # TODO: can we safely figure out how to use a pair of options
         # like these, adapted to as many platforms as possible, to use
         # multiple processors when available?
-        # extra_compile_args=['-fopenmp'],
+        extra_compile_args=['-ffloat-store'],
         # extra_link_args=['-fopenmp'],
```
brandon-rhodes, line 33: Could you keep the `openmp` compile arg next to the comment too? If I ever return to the attempt to do parallel processing, both those lines will need to be uncommented together. Thanks!

.github/workflows/test_package.yml
```diff
@@ -0,0 +1,191 @@
...
+
+
+  build_test_wheels:
```
brandon-rhodes, line 103: Could you remind me of the reason that this does not have the same `if` clause that `upload_to_pypi` does? I think I would be happy with normal PR's and commits just running tests against Python versions, if that could skip building the wheels.

brandon-rhodes (commented):
Thanks! I've asked a few questions, but overall I find this readable and I think I can tell what's going on, so I think we're close to merging!

.github/workflows/test_package.yml
mworion: Idea of the test strategy was:
Build dist and run tests in the repository -> always not accelerated. Therefore I added the line for checking if we really tested the unaccelerated version.

Build wheels and run test in the shell of cibuildwheels in explicit different directory -> accelerated.  
See explanation from cibuildwheels:
```
Shell command to run tests after the build. The wheel will be installed automatically and available for import from the tests. To ensure the wheel is imported by your tests (instead of your source copy), tests are run from a different directory.
````
Pretty sure that this works out as we discovered the 32bit precision issue with these tests reproducible.

If we keep it that way I even could remove the the print(accelerated) line for the first test. If we want to reduce CPU time I have to refactor the test from cibuildwheels to Github Actions directly. This might be cheaper as we don't need the docker image handling at all.

Let's decide which way to go.

sgp4/tests.py
mworion: Great, we keep that. Thanks for the hint.

setup.py
mworion: Sure, my fault. Should have kept that in.

.github/workflows/test_package.yml
mworion: See above: actually python testing does only test the unaccelerated version. As I wanted to test the accelerated version as well, I build the wheels and test in the cibuildwheel docker. To be decided about the strategy.

brandon-rhodes:
Thanks for the responses! I'm happy to go ahead and merge as soon as you replace that missing commented-out openmp option; that way, the project immediately has working tests again, and if I do decide after a few weeks I'd like a faster approach on master that skips building the wheels, then we can tackle that as a separate improvement. But let's see how things work with the approach already here — it's possible folks just accept the 15-minute turn-around time with no big problems, and likely that we never see high enough volume that we worry we're being expensive citizens of the ecosystem.

mworion:
Added back the comment line for compiler flags.

brandon-rhodes:
Thanks again for all your work on this! I look forward to using it and seeing its results posted to GitHub.

mworion:
You're welcome. If I could help with my limited capabilities.
Michel

brandon-rhodes:
@mworion — I thought you'd like to know that I just made the first successful release using GitHub Actions! I added the secret that the build needed, and the package files got uploaded:

https://pypi.org/project/sgp4/2.16/#files

At some point I'll need to figure out why 3.9 packages only got generated for Windows (maybe that's a cibuildtools limitation at this point?), but it looks like all three operating systems got covered for the earlier (and hopefully more popular) versions of Python 3!

mworion:
@brandon-rhodes 
Good to hear and thanks for the info. We might to have to look up, which python / os combination is actual valid for actual cibuildwheels. 
Michel

PR #103, opened 2022-05-02
Update CI and add support for more wheels

mworion:
Hi Brandon,

I made some updates for the wheel builds to platforms which were initially not feasible.  

Support Mac M1
Support ARM aarch64
Updates for used actions to recent ones and python runtime to 3.10

In addition I removed python 2.7 support for test (please feel free to add again)
I made some test trials and in my environment and local GitHub fork it went fine and build all wheels.

Hope this helps to get a broader usage (at least as aarch64 was missing and I had several users for which I build separately the wheels manually).

best regards,
Michel

### DIFF
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -44,23 +44,23 @@ jobs:
   build_test_dist:
     # Build the distribution in a matrix. Jobs are done in parallel.
     # The formal distro which is uploaded to PyPI will be build on
-    # ubuntu-latest for python 3.9, which is the recent one.
+    # ubuntu-latest for python 3.10, which is the recent one.
     # As in dist build no compilation takes place, we run all tests
     # in not accelerated mode.
 
     runs-on: ${{ matrix.os }}
     strategy:
-      max-parallel: 18
+      max-parallel: 15
       matrix:
-        python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


mworion:
I really don't know why the first check did fail with the following error message (which seems strange to me as the constraints are fulfilled anyway). Just adding a second commit without change to code (just adding an empty line) made the second run fine (as it happened during testing phase all the time)

Michel
```
The conflict is caused by:
[517](https://github.com/brandon-rhodes/python-sgp4/runs/6262777899?check_suite_focus=true#step:7:517)
      importlib-metadata 4.11.3 depends on zipp>=0.5
[518](https://github.com/brandon-rhodes/python-sgp4/runs/6262777899?check_suite_focus=true#step:7:518)
      The user requested (constraint) zipp==3.8.0
```

mworion:
Hello Brandon,
I don't want to bother you. As I want to avoid separately building wheels for aarch64 for my app users when I start to migrate to allow python 3.10 as well, is it possible that you have a look to the extension of the CI build and the PR I made ?
Many thanks in advance.
Michel

brandon-rhodes:
Thanks for your patience, I was at PyCon (my first in-person conference in three years!) when you first opened the issue, and after returning home I still haven't made it all the way through my backlog.

I don't have any objection to landing this, and I appreciate all the work that went into it—you've updated several things, and added several conditions to support the new platform.

Could you add 2.7 back in? With that change, I'll be happy to land it!

brandon-rhodes:
Thank you! I'll try bumping `release` to include this new code, and we'll see if some new wheels appear.

mworion:
Hi Brandon, thanks for moving forward. One remark: both windows and macOS raise a deprecation warning because of python 2.7 (I think that was the reason why I removed it), ubuntu does not. So far the GitHub action runs. It might be at a point in time when this fails if they switch from warning to error.
Michel

brandon-rhodes:
Thanks for the warning! We'll see how long the support continues.

PR #104, opened 2022-05-24
Convert MEAN_MOTION_DDOT to the correct unit when importing from OMM

signaleleven:


brandon-rhodes:
Well—drat! Now that I have a chance to look at this, I notice that the satellite used in most tests, Vanguard 1, has a `MEAN_MOTION_DDOT` of zero, which means that the presence or absence of a conversion factor is invisible to the tests!

I wonder if I should choose another satellite? Or, just add another value as a special case? I'll look at the options for making this visible to the tests!

PR #126, opened 2023-11-11
correct version number publish action

mworion:
@brandon-rhodes: It seems that publish action does not support just major versions numbers as others, so I choose a dedicated one like in the past.
As I cannot check with the correct credentials, there is the intension on PyPI to move away from User / password to certificates or authorization through tokens. I do not know, when this will be enforced. If so, we have to change the authorization method.
Hope this was a quick fix.

### DIFF
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -231,7 +231,7 @@ jobs:
     # chosen for the PyPI token.
 
     - name: upload_to_pypi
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
It worked! 🙂

mworion:
@brandon-rhodes: you're welcome. I use the lib together with sky field in my project as well and I'm happy to support.

PR #129, opened 2024-03-11
Add parse_json() function to omm

CDWimmer:
Added a test case also following the example set by the xml and csv tests - I think it's ready for review

### DIFF
--- a/sgp4/omm.py
+++ b/sgp4/omm.py
@@ -1,6 +1,7 @@
 """Support for the new Orbit Mean-Elements Message format for TLE data."""
 
 import csv
+import json
 import xml.etree.ElementTree as ET
 from datetime import datetime
 from math import pi
@@ -20,6 +21,13 @@ def parse_xml(file):
         for element in metadata, meanElements, tleParameters:
             fields.update((field.tag, field.text) for field in element)
         yield fields
+
+def parse_json(file):
+    json_data = json.load(file)
+    fields = {}
+    for item in json_data:
+        fields.update((k, v) for k, v in item.items())
+    yield fields
 
 _epoch0 = datetime(1949, 12, 31)
 _to_radians = pi / 180.0
--- a/sgp4/tests.py
+++ b/sgp4/tests.py
@@ -792,6 +792,30 @@ OBJECT_NAME,OBJECT_ID,EPOCH,MEAN_MOTION,ECCENTRICITY,INCLINATION,RA_OF_ASC_NODE,
 MARIO,1998-067UQ,2023-04-25T10:45:30.642912,15.99081912,.0014649,51.6242,216.2930,331.8976,28.1241,0,U,55123,999,1839,.1568E-2,.787702E-2,.29408E-3
 """
 
+MARIO_JSON = """\
+[
+    {
+        "OBJECT_NAME": "MARIO",
+        "OBJECT_ID": "1998-067UQ",
+        "EPOCH": "2023-04-25T10:45:30.642912",
+        "MEAN_MOTION": 15.99081912,
+        "ECCENTRICITY": 0.0014649,
+        "INCLINATION": 51.6242,
+        "RA_OF_ASC_NODE": 216.2930,
+        "ARG_OF_PERICENTER": 331.8976,
+        "MEAN_ANOMALY": 28.1241,
+        "EPHEMERIS_TYPE": 0,
+        "CLASSIFICATION_TYPE": "U",
+        "NORAD_CAT_ID": 55123,
+        "ELEMENT_SET_NO": 999,
+        "REV_AT_EPOCH": 1839,
+        "BSTAR": 0.001568,
+        "MEAN_MOTION_DOT": 0.00787702,
+        "MEAN_MOTION_DDOT": 0.00029408
+    }
+]
+"""
+
 def test_omm_xml_matches_old_tle():
     line0, line1, line2 = MARIO_TLE.splitlines()
     sat1 = Satrec.twoline2rv(line1, line2)
@@ -807,6 +831,16 @@ def test_omm_csv_matches_old_tle():
     sat1 = Satrec.twoline2rv(line1, line2)
 
     fields = next(omm.parse_csv(StringIO(MARIO_CSV)))
+    sat2 = Satrec()
+    omm.initialize(sat2, fields)
+
+    assert_satellites_match(sat1, sat2)
+
+def test_omm_json_matches_old_tle():
+    line0, line1, line2 = MARIO_TLE.splitlines()
+    sat1 = Satrec.twoline2rv(line1, line2)
+
+    fields = next(omm.parse_json(StringIO(MARIO_JSON)))
     sat2 = Satrec()
     omm.initialize(sat2, fields)
 


sgp4/omm.py
```diff
@@ -21,6 +22,13 @@ def parse_xml(file):
             fields.update((field.tag, field.text) for field in element)
         yield fields
 
+def parse_json(file):
```
brandon-rhodes, line 25: Happily, I think we can have users call `json.load()` directly. It looks like here you take the dictionary returned by `json.load()`, and then build a second identical dictionary, and then throw the first one away, which I think users might find a little slower than just using the original dictionary?

See if you can update this to make no change to `omm.py`, but instead make the test `test_omm_json_matches_old_tle()` just call `json.load()` itself. If it works and passes, then after this lands, I'll update the docs to show the same maneuver.

brandon-rhodes: Alternately, since I apparently chose to alias the function `csv.DictReader` as `parse_csv` in this file, we could do something like `parse_json = json.load`. But I suspect that we should simply avoid hiding the lower-level routines, in case folks need to use their options for things like encodings.

CDWimmer: This response accidentally became 3 times longer than I would have liked, I can only apologise: 

> It looks like here you take the dictionary returned by `json.load()`, and then build a second identical dictionary, and then throw the first one away, which I think users might find a little slower than just using the original dictionary?

Good spot, I think this thought was trying to wrestle its way into my head earlier today! 

However, I do believe wrapping it in a def provides a benefit:- your `parse_xml()` produces a generator rather than a `list` as would be returned by a raw `json.load`, so it will have consistent output regardless of input. A correction would be
```py
def parse_json(file):
    yield json.load(file)
```
to achieve the effect without throwing anything away.

A few more thoughts I've had materialise:
1. `json.load` takes a bunch of extra arguments that I don't think are very helpful here, and it would be good to not pollute users' IDEs with things that don't make sense to be present in this context. 

2. `json.load/loads()` has the annoying (for us) property that it will convert whatever JSON is entered into its Python equivalent _without complaint_, meaning were a user to not wrap their satellite data object in an array, they would receive no error relating to the totally valid JSON, but get an unexpected return type from `json.load()` -> a plain `dict`. ([see](https://docs.python.org/3/library/json.html#json.JSONDecoder)) One could build functionality into this to ensure a consistent return type. 

3. A fairly important note I think: I'm not sure either of us have used `yield` properly? I think we need to move `yield` inside the `for` block to actually take advantage of it, then construct the return value on the fly instead of building a list, e.g.  
```py
def parse_json(file):
    json_data = json.load(file)
    for item in json_data:
        yield {k: v for k, v in item.items()}
```
This way, we keep generator behaviour, don't throw away any data, and only generate a single `dict` at a time while still allowing for `for ... in parse(json)` usage. 
It's slightly slower for a handful of datasets, but almost 200,000 lines of [JSON](https://celestrak.org/NORAD/elements/gp.php?GROUP=active&FORMAT=json) relating to tens of thousands of satellites that is only increasing, it'll be vastly more memory efficient! Who knows what devices will be running this stuff after all. 
I am sure similar can be sought from parse_xml but I am actually falling asleep :( 

4. Getting slightly off-topic:
It could be made to accept both a filepath/object (i.e. stringIO) *and* a raw JSON string with a simple use of `isinstance()` and swapping to `json.loads()` (load _s_tring), or internally re-wrapping with stringIO, not sure on best approach. Same could be applied to the xml an csv parsers - because lord knows users will find incredible new ways to get their data into memory (I got here due to wanting to use a database!) 

5. 
> since I apparently chose to alias the function `csv.DictReader` as `parse_csv` in this file...

I'm tempted to suggest that normalising this to also produce a generator would be good :P

CDWimmer: Coming back to this with an almost reasonable amount of sleep had: I think it just makes more sense to return a list rather than use `yield` in all parse functions, I believe that I was wrong and there's no memory saved (will analyse that statement properly later if able). 
The entire active satellite JSON array comes to about 3.7MB. Even with added python overheads, if whatever is running an application can't handle that then it probably can't handle the rest of the application either; consequently imo it makes the most sense to just include a simple json.load wrapper for the convenience/consistency discussed prior, and swap the `yield`s for `return`s :)

ckuethe: I get just under 4MB of "compact" json (`jq -c`) from the combined active and analyst catalogs. Doing some quick'n'dirty tests, it seems that on my particular machine (linux x86_64), with my particular build of python (3.10.2), loading orbital elements takes about 1.15MB per thousand records. Let's maybe bump that up to 1.25MB per thousand records to accommodate all objects having non-empty `OBJECT_ID` and descriptive `OBJECT_NAME`. I don't have any 32-bit machines handy to test with.

This would be easier if the catalog was "ndjson" and you could have a generator that yields a line at a time which then gets parsed. Alas, I know of no existing incremental/streaming JSON parser that would return one dict at a time from a list of dicts..

CDWimmer: Results are in:
<details><summary>(click to show)</summary>
<p>

```
Averages over 100 test runs.
Running CSV test
Current memory usage is 4.224KB; Peak was 48.253KB; Diff = 44.029KB
CSV avg time:	0.1096 s
Running XML test
Current memory usage is 29.043KB; Peak was 38785.265KB; Diff = 38756.222KB
XML avg time:	0.8221 s
Running JSON generator test
Current memory usage is 7.952KB; Peak was 12533.226KB; Diff = 12525.274KB
JSON g avg time:	0.2179 s
Running JSON list test
Current memory usage is 7.52KB; Peak was 12532.538KB; Diff = 12525.018KB
JSON l avg time:	0.1969 s
```

</p>
</details> 

---
"There's a certain kind of programmer who'll look at any problem and say, 'I know, I will use XML.' Now he has two problems."

[Extremely bad test code here](https://gist.github.com/CDWimmer/a5e63d5066f99ecd49692fc11369ff5e)

To save you reading the code: it iterates over the entire active satellite dataset using each parser 100 times and averages the time taken. do the file opening beforehand (though it is on a drive with ~5GB/s reads so it wouldn't be much of a contributor. Running on an AMD Ryzen 7 5800X 8-Core under Win10. 

So, CSV's performance was a big surprise to me! It might be worth mentioning the speed benefits of using it somewhere, assuming I've done a fair test? 
Meanwhile, god only knows why people (i.e. CCSDS) keep writing standards that use XML, it sucked when it was invented and it sucks now. And more to the point, a generator for the JSON now verifiably doesn't help due to the lack of "streamability" of the format, so I concede just returning a list from `json.load()` is good enough - though I still think it could do with checking the resultant object is a `list` and not a `dict`.

File/download sizes are also very varied: 
csv - 1,342 KB
xml - 9,072 KB
json - 3,821 KB

ckuethe: Y'know what's terrible and works (given some bold assumptions):
```python
import os
import re
import mmap
import time
import json

def json_generator(fn):
    with open(fn, "r+b") as ifd:
        with mmap.mmap(ifd.fileno(), 0) as mm:
            for m in re.finditer(b'([{].+?[}])', mm, re.S|re.M):
                yield json.loads(m.group(0).decode())

test_single = False
data_file = "active.json"
if test_single:
    t = time.time()
    sat = json_generator(data_file).__next__()
    d = time.time() - t
    n = "1 record"
else:
    t = time.time()
    sats = list(json_generator(data_file))
    d = time.time() - t
    sat = sats[-1]
    n = f"{len(sats)} records"
print(f"loaded {n} in {d*1000:.2f}ms")
print(sat["OBJECT_NAME"])
```

but... it works:

```sh
$ ./awful.py # test_single = False
loaded 9452 records in 67.07ms
2024-047A
$ ./awful.py # test_single = True
loaded 1 record in 0.12ms
CALSPHERE 1
```

Loading pretty_printed records is less than 10ms slower to load the whole file (on my machine).

CDWimmer: I think I agree about csv, ultimately it is very predictable tabular data, of course it should be in a table! So it makes perfect sense to use pandas DataFrames too. 

Though, I don't know how consistent pandas has been across major Python revisions - this library supports 2.6 but the latest version of pandas requires >=3.9, so worth checking if there's version-specific caveats for examples/documentation. (Or release a final 2.X/3.5 version and then drop support, I am begging for type annotations 😢 3.5 is 9 years old!)

Generally I think docs should be an outline of "you can use this set of built in parsers provided by the library that are guaranteed to work with CelesTrak data (examples) [...] or if you want more advanced control of the data you can construct your own parser that has output in XYZ format... (pandas example that uses some DataFrame feature)"

ckuethe: > Well, that sure is cursed, but if it works... Run it through tracemalloc? I'm now emotionally invested in beating the CSV parser

"Your Scientists Were So Preoccupied With Whether Or Not They Could, They Didn’t Stop To Think If They Should"

I doubt that json loading will ever be faster than csv, but it was fun to see if I could do it without reading the whole file in.

brandon-rhodes: > I don't know how consistent pandas has been across major Python revisions…

Do you mean its performance? I have found that, if I'm careful, I can get it to switch into pure-C mode for CSV parsing in all versions that I try. Let me know if you hit a snag where it tries using Python parsing, and I'll try finding a workaround for you. Note that people using old Python versions usually just download and install the Pandas version that was current at that time, and can use Pandas just fine as long as they don't rely on recent tweaks and features.

> …this library supports 2.6

Just to make things simple, I've dropped the "2.6" from `setup.py` and from now on plan to only advertise 2.7 and later. I think it's fair at this point to tell Python 2.6 users to consider themselves permanent users of Skyfield 1.48.

CDWimmer: > Do you mean its performance? 

Sorry, I was unclear, I just mean if there have been changes to the "api", how things are used, feature changes etc. 

> install the Pandas version that was current at that time

Of course yeah :) Just worried if there were changes to how pandas does things then documentation examples might want to reflect that, but an excercise for the reader never hurt ;)

> I can get it to switch into pure-C mode for CSV parsing

That is great news about the C code, will see if I can get a quick something to do that, would you just `df = load_csv()` and then do `for row in df.iterrows()` as a fair(ish) comparison? Also how do you tell when it's doing pure-C? Just the time it takes? 

> only advertise 2.7

Sounds good, though I did read _a_ survey suggesting only about 1% or so of Python users were using 2.x now, though I suspect some physics/astronomy people aren't going to follow suite with that stat, do you know if it's known what the userbase of this is like?

brandon-rhodes:
I was busy with other projects for a few months, but, returning to this, I still don't find myself prepared to maintain new code in the project at this point. But we can still provide some sample code to help newer programmers avoid any struggle they might have in figuring out how to glue together the Python `json` module with `sgp4`. So I've landed an update to the docs:

f6cf651830829752923a9e109475424691232f83

As you can see, with the idea that users will want working code they have used, I have taken away the trick that I was doing with `next()` to pull only the first satellite, and instead illustrated a full `for` loop visiting every satellite in the file, and I have made this switch with all three OMM formats that are now documented.

I also put CSV first, JSON second, and XML in last place, to reflect what we learned as you were investigating.

I'm going to close this PR for now, but if I ever decide to add a JSON function, I'll re-open it. Thanks for sharing, I think that it did drive a good improvement to the documentation!

PR #134, opened 2025-03-03
fix missing revnum attribute on pure-python Satrec

plentydone:
Failure looks like this:

```
ERROR: test_export_tle_sgp4init (sgp4.tests.TestFunctions.test_export_tle_sgp4init)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/andrew/python-sgp4/.tox/py312/lib/python3.12/site-packages/sgp4/tests.py", line 252, in test_export_tle_sgp4init
    outline1, outline2 = export_tle(sat)
                         ^^^^^^^^^^^^^^^
  File "/home/andrew/python-sgp4/.tox/py312/lib/python3.12/site-packages/sgp4/exporter.py", line 89, in export_tle
    append(str(satrec.revnum).rjust(5))
               ^^^^^^^^^^^^^
AttributeError: 'Satrec' object has no attribute 'revnum'. Did you mean: 'elnum'?
```

Let me know if any adjustments are needed.


sgp4/model.py
```diff
@@ -89,6 +89,7 @@ def sgp4init(self, whichconst, opsmode, satnum, epoch, bstar,
 
         self.jdsatepoch = whole_jd
         self.jdsatepochF = fraction
+        self.revnum = 0
```
brandon-rhodes, line 92: Could you move this down to just beneath `self.classification`? That's what it sits next to over in the C++ code, and would make it a bit easier for my eyes to find it later than up here with the dates.

sgp4/tests.py
```diff
@@ -243,6 +243,17 @@ def test_tle_export():
...
+def test_export_tle_sgp4init():
+    sat = model.Satrec()
+    sat.sgp4init(
+        WGS84, 'i', VANGUARD_ATTRS['satnum'], VANGUARD_EPOCH,
+        *sgp4init_args(VANGUARD_ATTRS.copy())
+    )
+    outline1, outline2 = export_tle(sat)
+    parsed_sat = Satrec.twoline2rv(outline1, outline2, WGS84)
+    for attr in VANGUARD_ATTRS.keys():
+        assertEqual(getattr(parsed_sat, attr), getattr(sat, attr))
```
brandon-rhodes, line 255: I like how you didn't just make sure that `export_tle()` worked, but that the same satellite came back when re-parsed.

brandon-rhodes (changes requested):
Thanks for putting together this fix, complete with a test! It looks like a couple of tests are failing in CI, but I'll be happy to merge this once you get them fixed. Thanks!

sgp4/tests.py
plentydone: Thanks! Updated to use `assert_satellites_match` which I hadn't noticed before.

sgp4/model.py
plentydone: I think setting this to 0 here was a flawed approach, which I didn't see until I saw the failing CI tests. To make the patch less invasive, I adjusted my approach just to fix the error when exporting a TLE from a pure python Satrec initialized with sgp4init, as I'm not aware of anything else that needs it.

plentydone:
Sorry for the spam with the broken tests. I couldn't replicate locally for some reason, will figure out in the next day or two.

plentydone:
Finally came back to this, saw what I did wrong in the tests, and fixed and expanded them. Hopefully it's all good this time!

brandon-rhodes:
I'm hoping to find time next week for a release, to get this feature out. Thanks again for pointing out the problem.

PR #138, opened 2025-04-12
Adding build part support for python 3.13 aarch64 wheels

mworion:
Hi Brandon,
I was already on the way adding the necessary topics for the ci with python 3.13. You already started doing so. The pull request adds the missing parts.

- aarch64 needs again a split between python <= 3.12 and 3.13 ongoing due to the fact that the related numpy versions are not available for python 3.13

Looking forward to your comments. And hopefully you were able to publish a python 3.13 compatible version.

Michel

PS: I'm not happy with splitting the build even more, but I did not get a good solution otherwise -> still to need learning on this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,9 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     strategy:
-      max-parallel: 15
+      max-parallel: 18
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
> Looking forward to your comments. And hopefully you were able to publish a python 3.13 compatible version.

On the contrary, my attempt didn't work at all! Over in another tab, I have the recent build open, so I can try to figure out why my approach didn't work. I look forward to reading over this PR (maybe tomorrow there will be time) to see what I was missing. Thanks!

Issue #139, opened 2025-04-21
manylinux aarch64 wheels are not published


plentydone:
There are no manylinux-aarch64 wheels in pypi for this project, which would be very convenient for those (unfortunately) developing in Docker for Mac in arm64 containers. I see that CI is actually building these wheels, but I suspect their artifacts are being overwritten by the later x86_64 ubuntu build in the same workflow. I'm not very familiar with github actions but would be willing to try sending you a patch if you're interested.

mworion:
Hi, thanks for the hint. If you have a patch in mind, please let me know. An overwrite I would exclude as the naming is different, but you're right I do not use them, but in the pypi they are missing. I will check this tonight and look what's going on.
Michel

mworion:
Could you check if the wheels are ok in your setup ?

[wheelhouse-macos-13.zip](https://github.com/user-attachments/files/19895290/wheelhouse-macos-13.zip)
[wheelhouse-macos-14.zip](https://github.com/user-attachments/files/19895291/wheelhouse-macos-14.zip)
[wheelhouse-ubuntu-24.04-arm.zip](https://github.com/user-attachments/files/19895287/wheelhouse-ubuntu-24.04-arm.zip)
[wheelhouse-ubuntu-latest.zip](https://github.com/user-attachments/files/19895289/wheelhouse-ubuntu-latest.zip)
[wheelhouse-windows-latest.zip](https://github.com/user-attachments/files/19895288/wheelhouse-windows-latest.zip)

mworion:
I did a rework overall and made a PR for that.
Michel

brandon-rhodes:
The PR has landed and I ran a new `release` build. I now see files with names like this:

    sgp4-2.24-cp313-cp313-manylinux_2_17_aarch64.manylinux2014_aarch64.whl

— when I visit https://pypi.org/project/sgp4/#files and browse the file list. Hopefully those are the sorts of files that this issue was requesting? In the hopes that the answer is 'yes', I'm going to close this issue now, but feel free to comment here further, even on the closed issue, if a file is missing that you were expecting.

PR #140, opened 2025-04-24
Rework of CI

mworion:
Hi Brandon,

I do not want to bother you again with the CI build. I was still not satisfied about the GitHub build action solution at all. We got in addition #139 which prohibited the linux arch64 wheels. I decided to do a rework.
At the end, the workflow now has 100 lines less, and makes the next python release 3.14 done by just adding the string '3.14' in only 2 lines. Hopefully it is even more understandable. 

The wheel setup / version and the way of offering the MacOS solutions changed, but should be OK in PyPI packet index. 

Please review it at your convenience and please come back if changes are necessary.

Michel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@
 name: ci
 
 on:
-  # the trigger event for running the action is either a push on master or
+  # The trigger event for running the action is either a push on master or
   # release branch
 
   push:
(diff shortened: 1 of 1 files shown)


mworion:
Interesting, the build failed because of an HTTP gateway error. Never seen this before in my setup and is possibly an issue of GitHub.

mworion:
I decided to close this pull request an do another one to check, why the GitHub CI failed.

PR #141, opened 2025-04-25
Rework CI

mworion:
Hi Brandon,

I do not want to bother you again with the CI build. I was still not satisfied about the GitHub build action solution at all. We got in addition https://github.com/brandon-rhodes/python-sgp4/issues/139 which prohibited the linux arch64 wheels. I decided to do a rework.
At the end, the workflow now has 100 lines less, and makes the next python release 3.14 done by just adding the string '3.14' in only 2 lines. Hopefully it is even more understandable.

The wheel setup / version and the way of offering the MacOS solutions changed, but should be OK in PyPI packet index.

Please review it at your convenience and please come back if changes are necessary.

This is the second attempt without any changes to the first one to check if Github runners fail structurally what could not see in my local dev environment (also with GitHub)

Michel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@
 name: ci
 
 on:
-  # the trigger event for running the action is either a push on master or
+  # The trigger event for running the action is either a push on master or
   # release branch
 
   push:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,16 +45,17 @@ jobs:
   build_test_dist:
     # Build the distribution in a matrix. Jobs are done in parallel.
     # The formal distro which is uploaded to PyPI will be built on
-    # ubuntu-latest for python 3.12.
+    # ubuntu-latest for python 3.13.
     # As in dist build no compilation takes place, we run all tests
     # in not accelerated mode.
 
     runs-on: ${{ matrix.os }}
+    name: Build dist ${{ matrix.python-version }} on ${{ matrix.os }}
     strategy:
-      max-parallel: 18
+      max-parallel: 24
       matrix:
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
> I do not want to bother you again with the CI build.

I am not bothered at all; thank you for this additional work, which, I agree, makes the CI steps easier to read and reason about.

> I was still not satisfied about the GitHub build action solution at all. We got in addition #139 which prohibited the linux arch64 wheels. I decided to do a rework. At the end, the workflow now has 100 lines less, and makes the next python release 3.14 done by just adding the string '3.14' in only 2 lines. Hopefully it is even more understandable.

Reading through it, it does look easier to understand! Now that it's landed, I will wait until the next time I need to do a release; then, we'll see how it works, and whether everything makes it to PyPI successfully, and whether I really understand how it's working or not. 🙂

Thanks again for the contributions you make to the project!

brandon-rhodes:
@mworion — I tried bumping the `release` branch to this new code, but the final step ran into an error:

https://github.com/brandon-rhodes/python-sgp4/actions/runs/14810389138/job/41584454227

Looking at the diff, I understand the old code, which did a separate `download-artifact` on each of the products that had been uploaded in the sections above it. But I'm not sure how the new code was supposed to work that just asked for a single named artifact? Is there a quick tweak that will get it working? Thanks for taking a look!

mworion:
@brandon-rhodes: I'm so very sorry, I overlooked the breaking change from upload/download artifacts v3 -> v4. I mentioned that I started migrating to 3.13 , but I did it on v3. You already started with v4 (which was good) and I merged it on my side, but did not notice it.
So I corrected this. I will submit a new PR again the current master.
Michel

PR #142, opened 2025-05-04
fix breaking change in artifact handling v3 to v4

mworion:
@brandon-rhodes,
Attached the fix for the broken part in artefactory workflow. I did try to test is as far I could go (I made a release branch like yours to see if the workflow is running without problems. I could not check the upload itself, but the files are now present in dist folder. So it should run through now.
I also removed the upgrade pip installation as it does not remove the single warning in workflow. This happens during the python-setup workflow, which is imported.
Again sorry for the missing changes I forgot.
Michel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,9 +70,7 @@ jobs:
       # testing and checking the metadata.
 
       - name: install dependencies
-        run: |
-          python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
>Again sorry for the missing changes I forgot.

No problem — it cost me only a few minutes of work to test a release and then circle back to you with the error message — but you have saved me _far_ more than a few minutes by volunteering to help with the build, so thanks very much!

PR #144, opened 2025-06-26
Adding support for pandas.DatetimeIndex to jday_datetime()

WeatherGod:


brandon-rhodes:
Good morning, @WeatherGod!

My first impulse is to keep sgp4 as simple as possible, and to treat is as a 'leaf node' in the great dependency tree of life, such that it knows nothing about other Python libraries. While a small forest of type checks would indeed make things more convenient for the users of other libraries, I'm not sure where I'd draw the line when folks with other third-party libraries wanted to add type checks for them as well?

If we were going to add this to sgp4, I would probably instead design an `sgp4.pandas` submodule that had wrappers around the standard sgp4 objects, that did the conversions. That way, _n_ third-party libraries could eventually be supported, but without any `if` statements, and where users would opt-in to the additional code by importing it optionally; users who didn't need Pandas type support would simply not import `sgp4.pandas`. But, again, I think sgp4 will probably have a happier future if it's kept ignorant of other libraries and their conventions?

Can we think of somewhere else that this might go, to solve the same problem without adding more code to what I hope to keep as a very slim library? Feel free to explain your situation in more detail!

WeatherGod:
The issue in the accelerated version of `sgp4_array()` is that it obtains `array = type(jd)`, which for a numpy array is `np.ndarray`, and for a pandas object is something like `pd.Index` or `pd.Series`. The problem with that is that the semantics are different for these constructors. `np.ndarray` expects a shape parameter, while just every single other array-like type I know expects data in their constructors.

The approach I would take here is to make `sgp4_array` be a ufunc. The beauty of ufuncs is that you are off-loading all of the responsibility of managing disparate array-like types to numpy. List of floats, pandas Series, dask arrays, etc., will all work so long as they implement the `__array_wrap__()` protocol. https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html. We can also utilize this approach for the `jday()` and `jday_datetime` functions, but either it would have to be `jday_array()`, or make numpy a dependency to have it available across the board.

WeatherGod:
Initial attempt at utilizing this feature, even at the pure-python level is encouraging, but still has problems with pandas datetimeindex.

brandon-rhodes:
I really like the idea of a ufunc, where the library gets out of the business of trying to do any broadcasting itself, and instead lets NumPy handle it all. I guess it wouldn't be quite as fast, because there would be a level of indirection involved, but my guess is the difference in speed would be only a couple of percent?

Instead of having this 'sgp4' library earn a NumPy dependency, we could create a separate package `sgp4_ufunc` or something that does require NumPy and packages SGP4 up solely for consumption by tools like NumPy and maybe Pandas too. It would be a much simpler library than this one, because it could drop the old fallback Python version of SGP4, and concentrate solely on wiring the SGP4 C++ code together with NumPy.

WeatherGod:
Another advantage of a brand-new package is we can clean up the interface without worrying about backwards compatibility. The problem though with new packages trying to replace widely used packages is uptake (and naming!). We could do this as an optional feature (e.g. - `pip install sgp4[array]`) that would include this in the package. With conda, I believe this means two "variants". Not quite sure, but the numpy one will need to be tied to the numpy build because we would have to do this ufunc in C.

Playing around with this a bit, at the moment I can't quite get the python-based ufunc stuff working because the return is a tuple of tuples, so it keeps thinking I'm trying to do object arrays. I'm pretty sure we can handle this properly in the C++ code instead, but I've never personally tried doing this. Before committing to anything, I'm going to have to try doing a dummy ufunc that takes inputs and outputs similarly and see what happens with pandas objects.

brandon-rhodes:
> So, I guess the question is now, how do we want to go forward from here?

My guess is that we would want to create a new project, maybe in a team like:

https://github.com/skyfielders

— rather than under my own personal GitHub account. We could call it `sgp4_ufunc`. I could easily toss in the official C++ code, and then maybe you could take charge of adding the little bit of code that's the ufunc itself? I would be happy to provide code review. Then, the big step would be creating a GitHub Actions command that compiled it for all platforms. Would you be willing to tackle that as well? I'm not sure I have the cycles in July to tackle it myself, but you could use the `sgp4` GitHub Action as a starting point.

Let me know if you're interested in moving forward with it, or if it sounds a bit daunting, even with my providing code review, to turn this into a Python package 🙂

WeatherGod:
I'm also spending a bit of time lifting the SGP4.cpp, wrapper.cpp, and SGP4.h files over. I'm thinking of dropping SatrecArray and just letting numpy handle it. The interesting question is whether or not we can turn Satrec into a numpy dtype (and, is it worth doing so versus just having a numpy array of objects).

brandon-rhodes:
> Alright! I managed to get a gufunc working that takes Satrec, jd, and fr (scalars or array-like), broadcasts them, and does the processing, returning e, r, and v. I also have a few rudimentary tests (mostly smoke tests).

Congratulations! It's a big step to have the basic concept all working, with the basic pieces wired together.

> Let me know when you are ready for the next steps and set up a repo for me to make a pull request with this initial set of stuff. There is still a lot to clean up, but the original vallado c++ code has been lifted into the package and a module created. I have not yet dealt with any IO or convenience functions or somesuch.

I have just set up a repo where you can start committing files! Maybe you can just directly commit what you have so far, and then we can switch to PRs once the basic package is there.

https://github.com/skyfielders/python-sgp4-ufunc/

I think the published package should probably, alas, avoid having a short name, and be called `sgp4_ufunc` because when folks see the name in a list of packages, they are going to have very little to go on beyond the name for guessing what it does; I don't think that seeing a lone letter `u` will let them guess the package's intent? It's taken me many years to go with longer names, because I so love being brief, but I find over the long term that support is much easier if big obvious names are chosen, even if they require more typing.

We can continue discussion over in the new repo, but I'll be curious to see: is the `sgp4` struct itself a plain Python type in your new code? Or, is it a NumPy `dtype`, which would let many satellites be put into a NumPy array without the overhead of needing _n_ Python objects for _n_ satellites? (Since lots of users of `sgp4` in fact simply need to instantiate the whole sky of satellites and then ask them all for positions, to see which satellite might be in a photograph, say.)

WeatherGod:
I get the point about the name, but still not sold on this name (most users won't know what a ufunc is and why it would be a good thing), but we can punt on that while we get this to a working state. I'll adjust and commit soon and we can iterate from there.

In the meantime, I'm getting concerned about something I just noticed in the Vallado code. It keeps modifying the Satrec structure. That makes the algorithm (1) not thread-safe, and (2) possibly not reorderable? This has implications in situations where someone is using a dask array, for example. Not sure if I should open an issue here, or in the new project for discussion.

brandon-rhodes:
I am going to go ahead and close this PR, as, per the above discussion, I'm not planning on adding pandas-specific code to this `sgp4` project. I look forward to further discussions over on your new repository, though!

PR #145, opened 2025-06-27
Speed up pure-python sgp4_array() by not growing a list

WeatherGod:


WeatherGod:
Noticed this while reviewing both `sgp4_array()` implementations. Seemed like a low-hanging fruit.

brandon-rhodes:
Interesting! I had thought that this line:

    results = [self.sgp4(jd_i, fr_i) for jd_i, fr_i in zip(jd, fr)]

— also created an empty list and started appending to it. Are you saying that a size hint is passed to the comprehension from `zip()`, which must peek at one or both inputs to learn ahead of time how many items it will see?

And, what kind of speedup are you seeing from this change?

WeatherGod:
Hmm, actually testing it out and I'm not seeing much discernible difference. In theory, the list comprehension is supposed to be faster. I don't know if there is some sort of size hinting being passed through or not, but in my experience this should always be a option for some performance boost.

Anyway, this may end up being moot anyway if you like my idea about adopting ufuncs everywhere.

WeatherGod:
It could very well be that with the many improvements elsewhere in python, the differences between a list comprehension and explicit `append()` has shrunk significantly.

brandon-rhodes:
I'm going to go ahead and close this; the `for` loop makes it easy to toss in a `print()` statement for debugging, which the comprehension prevents, so for now I'm inclined to keep it as-is.

PR #147, opened 2025-10-16
Python 3.14 support

scottshambaugh:
Building sgp4 with python 3.14 is working out of the box for me with clean tests. This runs it through CI.

(Though it looks like it needs maintainer approval to run)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       max-parallel: 24
       matrix:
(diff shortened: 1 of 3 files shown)


brandon-rhodes:
@scottshambaugh — Thanks for this update! I should have jumped in and answered you back in October, but I was hiking the Grand Canyon, alas, and then didn’t properly dig myself out from all the email that arrived.

@mworion — I've gone ahead and merged this PR, since Scott waited three months for my reply, and since CI is broken anyway so this doesn't do additional harm. I would welcome further tweaks to get CI working again, if you'd like to offer them!

PR #150, opened 2026-01-13
Update CI and add python 3.14 support

mworion:
@brandon-rhodes:
This pull request follows path 1) of our discussion. I added the python 3.14 support and updated the actions to the latest versions. 
The action is now shorter and the accelerated version are now tested against all python versions (was: only latest).
I added some comments with hints, what to be done when adding python 3.15 support.
Unfortunately some of the runners (ubuntu-arm and macos-intel) do not have a tag '-lastest', so also in future there should be a look to it.
Michel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,15 +45,17 @@ jobs:
   build_test_dist:
     # Build the distribution in a matrix. Jobs are done in parallel.
     # The formal distro which is uploaded to PyPI will be built on
-    # ubuntu-latest for python 3.13.
+    # ubuntu-latest for python 3.14.
     # As in dist build no compilation takes place, we run all tests
     # in not accelerated mode.
 
     runs-on: ${{ matrix.os }}
     name: Build dist ${{ matrix.python-version }} on ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


mworion:
@brandon-rhodes:
Did you already release the 3.14 updates? I still see the 08.2025 version 2.25 and no compiled 3.14 binaries.
I can install skyfield with SGP4 also on 3.14 (which should use it on pure python), but I get some differences -> this will be part of an issue I will create.
Many thanks in advance.
Michel

brandon-rhodes:
@mworion — Alas, I didn't make a release yet, did I? The series of pull requests that I merged at the beginning of the month caused a bit of breakage for my test scripts, which interrupted me, and I didn't get back to the project quickly.

I'll plan to get those scripts working again today, probably bump the project version number since at least the metadata will now change to proclaim 3.14, and try a release ­— the first time we'll do a live trial of an `sgp4` that doesn't need one binary per Python version. Fingers crossed that it will work well!

neutrinoceros:
Fingers crossed ! Please let me know if anything breaks (or if nothing does 😄)

mworion:
@brandon-rhode, @neutrinoceros 
many thanks for the contributions, looking forward to test !

PR #151, opened 2026-03-08
BLD/WHL: declare Python 3 wheels as abi3-compliant

neutrinoceros:
this may overlap partially with #150 (I'll notify the author there)
In short, this makes it so that all supported Python 3 versions (including future ones) are served with a single wheel (per platform). The tradeoff with this technique is usually that some faster C APIs may not be available, but it looks like the extension here isn't relying on any of those, so it's actually free here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,11 +14,26 @@
 
 import os
 import sys
+import sysconfig
 from distutils.ccompiler import get_default_compiler
 from distutils.core import setup, Extension
 
 PYTHON_SGP4_COMPILE = os.environ.get('PYTHON_SGP4_COMPILE', '')
+
+USE_PY_LIMITED_API = (
+    # Py_buffer is required
+    sys.version_info >= (3, 11)
+    # LIMITED_API is not compatible with free-threading (as of CPython 3.14)
+    and not sysconfig.get_config_var("Py_GIL_DISABLED")
+)
+ABI3_TARGET_VERSION = "".join(str(_) for _ in sys.version_info[:2])
+ABI3_TARGET_HEX = hex(sys.hexversion & 0xFFFF00F0)
+
 ext_modules = []
+define_macros = []
+
+if USE_PY_LIMITED_API:
+    define_macros.append(("Py_LIMITED_API", ABI3_TARGET_HEX))
 
 if sys.version_info[0] == 3 and PYTHON_SGP4_COMPILE != 'never':
 
@@ -44,6 +59,8 @@ if sys.version_info[0] == 3 and PYTHON_SGP4_COMPILE != 'never':
             'extension/wrapper.cpp',
         ],
         extra_compile_args=extra_compile_args,
+        define_macros=define_macros,
+        py_limited_api=USE_PY_LIMITED_API,
     ))
 
 # Read the package's docstring and "__version__" without importing it.
@@ -98,5 +115,8 @@ setup(
     package_data = {'sgp4': ['SGP4-VER.TLE', 'sample*', 'tcppver.out']},
     provides = ['sgp4'],
     ext_modules = ext_modules,
+    options={
+        "bdist_wheel": {"py_limited_api": "cp%s" % ABI3_TARGET_VERSION}
+    } if USE_PY_LIMITED_API else {},
     **setup_kwargs,
 )


extension/wrapper.cpp
```diff
@@ -1,3 +1,6 @@
+// keep in sync with setup.py setup(options=...)
+#define Py_LIMITED_API 0x030800f0
```
neutrinoceros: note that I didn't make this conditional because (supposedly) Python 2 isn't even aware of this pre-proc def so it shouldn't choke on it

(https://peps.python.org/pep-0384/)

neutrinoceros:
I've opened the following PRs in support of this patch:
- #153 
- #154 
- #156 
- #157 

the latter 3 should represent the *entire* set of changes required, and are all based of the first one so CI would fail if anything isn't working.
At the time of writing, #157 is the only part that does not compile locally for me yet.

neutrinoceros:
FWIW I successfully ran the test suite locally with all the preliminary PRs + this one combined locally, but I do think we should take care of #158 before merging anything beyond #153

neutrinoceros:
@brandon-rhodes all 3 required PRs are now stable and ready for review

setup.py
```diff
@@ -14,11 +14,26 @@
...
 PYTHON_SGP4_COMPILE = os.environ.get('PYTHON_SGP4_COMPILE', '')
+
+USE_PY_LIMITED_API = (
+    # Py_buffer is required
+    sys.version_info >= (3, 11)
+    # LIMITED_API is not compatible with free-threading (as of CPython 3.14)
+    and not sysconfig.get_config_var("Py_GIL_DISABLED")
+)
+ABI3_TARGET_VERSION = "".join(str(_) for _ in sys.version_info[:2])
+ABI3_TARGET_HEX = hex(sys.hexversion & 0xFFFF00F0)
```
brandon-rhodes, line 30: This is my first time seeing the process of building an `abi3` extension, so this question might not make sense, but: shouldn't we have something like 0x030b0000 here (assuming that means "Python 3.11"; I'm doing the binary in my head) regardless of which Python we happen to be using to build the binary wheel? Or must Python 3.11 be used for the compilation if the resulting binary wheel is going to be compatible that far back — does the ABI, in other words, only make a build forward-compatible to new Python versions, but not backwards to older ones?

neutrinoceros: > does the ABI, in other words, only make a build forward-compatible to new Python versions, but not backwards to older ones?

essentially, that's it.
For the longer version, see [this paragraph on the official docs](https://docs.python.org/3/c-api/stable.html#limited-api-caveats).

neutrinoceros: >  `0x030b0000` here (assuming that means "Python 3.11"; I'm doing the binary in my head)

that's almost correct (and in practice, it works), though to be precise it means "3.11.0 *any*" (that is, any alpha or beta version will also match). The final `F0` I'm using means "3.11.0 *final*", which excludes pre-releases of 3.11. Generally, the ABI is supposed to be stabilized on the first release candidate, but there have been (minor) exceptions to that rule in the past so I'm using "final" as an extra level of guarantee in stability.

brandon-rhodes: Thanks for the link to the documentation! I have also now searched GitHub with patterns like:

    Py_LIMITED_API /0x[0-9A-Fa-f]+/

It looks like most projects simply set the value to a literal hex number like `0x03080000` or `0x030A0000` depending on what version of the ABI they support — which shouldn't change merely depending on the version of Python they compile against, I don't think? If you accidentally compile using, say, 3.12 instead of 3.11, you don't get an extension that doesn't work with 3.11, I don't think? The language that the docs use is:

> For these reasons, we recommend testing an extension with all minor Python versions it supports, and preferably to build with the lowest such version.

Which makes compiling with, say, 3.11 instead of 3.12 a nice-to-have, but not a necessity, I don't think?

neutrinoceros: > most projects simply set the value to a literal hex number

yes, but that shortcut is technical debt in my opinion: it won't update itself when these projects drop Python versions, and might go out of sync with requirements.

> if you accidentally compile using, say, 3.12 instead of 3.11, you don't get an extension that doesn't work with 3.11, I don't think?

The result is undefined. You can't know for sure unless you test the wheel with an older version too. Indeed, most often it'll just work, but if you're going to test all versions anyway (which you do), this is the most maintainable way to do things IMHO.
I should mention, in case it's not obvious, that it deliberately compounds with how `cibuildwheel` works by default: it iterates through Python versions from oldest to newest, and will skip the build step, going straight to tests, if there's already a compatible artifact. If you look at this PR's logs, you'll see the following steps:
- build `cp310-cp310`
- test `cp310-cp310` w Python 3.10
- build `cp311-abi3`
- test `cp311-abi3` w Python 3.11
- test `cp311-abi3` w Python 3.12
- test `cp311-abi3` w Python 3.13
- test `cp311-abi3` w Python 3.14

brandon-rhodes: Thanks, I didn't know `cibuildwheel` didn't just do all the builds in parallel without interaction between them!

What I'll probably do is add a final testing step that (a) lists the artifacts, (b) verifies that there's an `abi3` artifact but no artifacts for the Python versions that the ABI version should cover, and (c) verifies that the `api3` wheel declares the hex that I'm expecting as its Limited API version.

That way, if I accidentally mangle the build such that it produces a version I'm not expecting, I'll be notified and it will fail, instead of releasing into the wild a wheel that doesn't support the users I intend.

But I think that needn't affect this PR! I'll re-read to be sure —

neutrinoceros: > What I'll probably do is add a final testing step that (a) lists the artifacts, (b) verifies that there's an abi3 artifact but no artifacts for the Python versions that the ABI version should cover, and (c) verifies that the api3 wheel declares the hex that I'm expecting as its Limited API version.

You could also use `abi3audit` for this, though [there's a discussion going on in cibuildwheel](https://github.com/pypa/cibuildwheel/issues/1342) to perform this automatically in the future

neutrinoceros: also, `cibuildwheel` already lists artifacts as a markdown table into `GITHUB_STEP_SUMMARY`. If you just want to make it more visible, you could start from there !

PR #154, opened 2026-03-11
RFC: redefine PyTypeObjects as heap types for Limited API compliance

neutrinoceros:
based off #153 
needed for #151

At the time of opening, I'm getting a single compilation error:
```
clang++: warning: optimization flag '-ffloat-store' is not supported [-Wignored-optimization-argument]
extension/wrapper.cpp:582:10: error: use of undeclared identifier 'Py_tp_as_sequence'
  582 |         {Py_tp_as_sequence, &SatrecArray_as_sequence},
```

I don't have experience with "sub slots", which [seem needed for defining the `tp_as_sequence` slot correctly](https://docs.python.org/3/c-api/typeobj.html#tp-slots). Hopefully I can crack it soon but in the mean time, any help would be appreciated.

### DIFF
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -22,6 +22,9 @@ typedef struct {
     elsetrec satrec[0];
 } SatrecArrayObject;
 
+static PyObject* SatrecType = NULL;
+static PyObject* SatrecArrayType = NULL;
+
 /* Support routine that is used to support NumPy array broadcasting for
    both individual satellite objects and also arrays. */
 
@@ -483,21 +486,12 @@ static PyGetSetDef Satrec_getset[] = {
     {NULL},
 };
 
-static PyTypeObject SatrecType = {
-    PyVarObject_HEAD_INIT(NULL, 0)
-    /* See the module initialization function at the bottom of this file. */
-};
-
 /* Details of the SatrecArray. */
 
 static Py_ssize_t
 Satrec_len(PyObject *self) {
     return ((SatrecArrayObject*)self)->ob_base.ob_size;
 }
-
-static PySequenceMethods SatrecArray_as_sequence = {
-    Satrec_len
-};
 
 static PyObject *
 SatrecArray_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
@@ -529,7 +523,7 @@ SatrecArray_init(SatrecArrayObject *self, PyObject *args, PyObject *kwds)
         PyObject *item = PySequence_GetItem(sequence, i);
         if (!item)
             return -1;
-        if (!PyObject_IsInstance(item, (PyObject*) &SatrecType)) {
+        if (!PyObject_IsInstance(item, (PyObject*) SatrecType)) {
             PyErr_Format(PyExc_ValueError, "every item must be a Satrec,"
                          " but element %d is: %R", i, item);
             Py_DECREF(item);
@@ -557,9 +551,40 @@ static PyMethodDef SatrecArray_methods[] = {
     {NULL, NULL}
 };
 
-static PyTypeObject SatrecArrayType = {
-    PyVarObject_HEAD_INIT(NULL, sizeof(elsetrec))
-    /* See the module initialization function at the bottom of this file. */
+static char doc_SatRecType[23] = "SGP4 satellite record.";
+static PyType_Slot SatrecType_slots[] = {
+    {Py_tp_doc, doc_SatRecType},
+    {Py_tp_methods, Satrec_methods},
+    {Py_tp_members, Satrec_members},
+    {Py_tp_getset, Satrec_getset},
+    {Py_tp_new, (void *)PyType_GenericNew},
+    {0, NULL},
+};
+
+static PyType_Spec SatrecType_spec = {
+    .name = "sgp4.vallado_cpp.Satrec",
+    .basicsize = sizeof(SatrecObject),
+    .itemsize = 0,
+    .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE | Py_TPFLAGS_IMMUTABLETYPE,
+    .slots = SatrecType_slots,
+};
+
+static char doc_SatRecArrayType[26] = "SGP4 array of satellites.";
+static PyType_Slot SatrecArrayType_slots[] = {
+    {Py_tp_doc, doc_SatRecArrayType},
+    {Py_tp_methods, SatrecArray_methods},
+    {Py_tp_init, (void *)SatrecArray_init},
+    {Py_tp_new, (void *)SatrecArray_new},
+    {Py_sq_length, (void *)Satrec_len},
+    {0, NULL},
+};
+
+static PyType_Spec SatrecArrayType_spec = {
+    .name = "sgp4.vallado_cpp.SatrecArray",
+    .basicsize = sizeof(SatrecArrayObject),
+    .itemsize = sizeof(elsetrec),
+    .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE | Py_TPFLAGS_IMMUTABLETYPE,
+    .slots = SatrecArrayType_slots,
 };
 
 /* The module that ties it all together. */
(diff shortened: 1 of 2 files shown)


neutrinoceros:
... which might not be a problem since, IIUC, the extension is only ever compiled for Python 3 anyway

neutrinoceros:
I got it to compile locally. Hopefully it runs smoothly in CI too.

neutrinoceros:
failure is real. I'll probably need to iterate a bit if I cannot reproduce this locally (which remains to be seen)

neutrinoceros:
I can reproduce, this shouldn't take long

neutrinoceros:
shotgun debugging: it seems that this is the block that errors out within the Limited API build
```C
        if (!PyObject_IsInstance(item, (PyObject*) &SatrecType)) {
            PyErr_Format(PyExc_ValueError, "every item must be a Satrec,"
                         " but element %d is: %R", i, item);
            Py_DECREF(item);
            return -1;
        }
```
from
`SatrecArray_init(SatrecArrayObject *self, PyObject *args, PyObject *kwds)`

extension/wrapper.cpp
```diff
@@ -529,7 +523,7 @@ SatrecArray_init(SatrecArrayObject *self, PyObject *args, PyObject *kwds)
         PyObject *item = PySequence_GetItem(sequence, i);
         if (!item)
             return -1;
-        if (!PyObject_IsInstance(item, (PyObject*) &SatrecType)) {
+        if (!PyObject_IsInstance(item, (PyObject*) SatrecType)) {
```
neutrinoceros, line 526: took me a couple minutes to spot my mistake, but it was actually a trivial fix

extension/wrapper.cpp
```diff
@@ -557,9 +551,40 @@ static PyMethodDef SatrecArray_methods[] = {
     {NULL, NULL}
 };
 
-static PyTypeObject SatrecArrayType = {
-    PyVarObject_HEAD_INIT(NULL, sizeof(elsetrec))
-    /* See the module initialization function at the bottom of this file. */
+static char doc_SatRecType[23] = "SGP4 satellite record.";
```
brandon-rhodes, line 554: Can this string simply be inlined into the struct? Or is there a limitation here that requires it to have its own name? (Also, I think `[23]` can be replaced with `[]` to make it easier on anyone who edits the string in the future.) (The same question applies also to the string farther down.)

extension/wrapper.cpp
```diff
@@ -575,45 +600,25 @@ static PyModuleDef module = {
...
-    SatrecArrayType.tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE;
-    SatrecArrayType.tp_doc = "SGP4 array of satellites.";
-    SatrecArrayType.tp_methods = SatrecArray_methods;
-    SatrecArrayType.tp_init = (initproc) SatrecArray_init;
-    SatrecArrayType.tp_new = SatrecArray_new;
-
-    if (PyType_Ready(&SatrecArrayType) < 0)
+    SatrecArrayType = PyType_FromSpec(&SatrecArrayType_spec);
+    if (SatrecArrayType == NULL)
         return NULL;
```
brandon-rhodes, line 610: Does this `return` need a `Py_DECREF(&SatrecType);` to be run just before returning?

extension/wrapper.cpp
```diff
@@ -575,45 +600,25 @@ static PyModuleDef module = {
...
 
     PyObject *m = PyModule_Create(&module);
     if (m == NULL)
         return NULL;
 
-    Py_INCREF(&SatrecType);
-    if (PyModule_AddObject(m, "Satrec", (PyObject *) &SatrecType) < 0) {
+    if (PyModule_AddObject(m, "Satrec", SatrecType) < 0) {
         Py_DECREF(&SatrecType);
```
brandon-rhodes: I think you also need a `Py_DECREF(&SatrecArrayType);` here, since otherwise you will leak that object by losing the only pointer to it?

setup.py
```diff
@@ -39,7 +40,10 @@
         # multiple processors when available?
         # extra_compile_args=['-fopenmp'],
         # extra_link_args=['-fopenmp'],
-        extra_compile_args=['-ffloat-store'],
+        extra_compile_args=[
+            '-ffloat-store',
+            "-std=c++20" if get_default_compiler() != "msvc" else "/std:c++20",
```
brandon-rhodes: Let's add a comment explaining why this is necessary, and why both compilers wouldn't choose the most recent C++ automatically.

brandon-rhodes (changes requested):
Thanks again for all of these PRs, which I'll try to make better progress through in May than I did in April. Here are a few questions about this one. Once they're answered, I think it will be ready to merge! This was interesting to read, as it's my first introduction to the new mechanisms for defining types.

extension/wrapper.cpp
neutrinoceros: I initially tried with it inlined but found that one compiler was struggling (maybe clang, but I don't recall the exact details), while giving it its own variable works on all targets exercised here.

setup.py
neutrinoceros: Sure. I'll probably need to revert this part just refresh the logs, because I don't recall the specifics now.

neutrinoceros:
Answering from my phone right now. 3/4 of your questions require my double-checking. I'll try to squeeze this into my afternoon work session !

extension/wrapper.cpp
neutrinoceros: I think that's correct. I also conclude that I can collapse two now-identical blocks.

extension/wrapper.cpp
neutrinoceros: agreed, thanks for spotting this

setup.py
neutrinoceros: see this log: https://github.com/brandon-rhodes/python-sgp4/actions/runs/25109895590/job/73581108136?pr=154
```
  extension/wrapper.cpp(22): warning C4200: nonstandard extension used: zero-sized array in struct/union
  extension/wrapper.cpp(22): note: This member will be ignored by a defaulted constructor or copy/move assignment operator
  extension/wrapper.cpp(202): warning C4244: 'initializing': conversion from 'long' to 'char', possible loss of data
  extension/wrapper.cpp(565): error C7555: use of designated initializers requires at least '/std:c++20'
  extension/wrapper.cpp(583): error C7555: use of designated initializers requires at least '/std:c++20'
  error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x86\\cl.exe' failed with exit code 2
```

I don't know why msvc doesn't pick c++20 on its own though.

neutrinoceros: (I added this line back on my latest push so that you *can* merge this if you want)

brandon-rhodes: 1. If `-std=c++20` doesn't seem to be needed, then it would be nice to adjust this so that it's not supplied; I'd rather not introduce arguments that aren't necessary.
2. If you could, adjust the `if` to be positive rather than negative; that's a Raymond Hettinger piece of advice (if I recall?) that does tend to make things more readable for me in the future.
3. Thanks for digging out those logs! I can add a comment after merging to explain to myself why we have to adjust the MS compile argument.

extension/wrapper.cpp
```diff
@@ -575,45 +600,22 @@ static PyModuleDef module = {
...
-    if (PyModule_AddObject(m, "Satrec", (PyObject *) &SatrecType) < 0) {
-        Py_DECREF(&SatrecType);
-        Py_DECREF(m);
-        return NULL;
-    }
-
-    Py_INCREF(&SatrecArrayType);
-    if (PyModule_AddObject(m, "SatrecArray", (PyObject *) &SatrecArrayType) < 0) {
+    if ((PyModule_AddObject(m, "Satrec", SatrecType) < 0)
+        || (PyModule_AddObject(m, "SatrecArray", SatrecArrayType) < 0)) {
```
brandon-rhodes: Alas, I think this introduces an error in the case that the first `PyModule_AddObject()` succeeds but the second fails: in that case, we wind up calling `Py_DECREF(&SatrecType);` with the result that `SatrecType` is freed, even though it's still a member of the module with the name `"Satrec"`. When the module is the finalized and discarded, I think it would experience a double-free?

extension/wrapper.cpp
```diff
@@ -575,45 +600,22 @@ static PyModuleDef module = {
...
         Py_DECREF(&SatrecArrayType);
         Py_DECREF(&SatrecType);
```
brandon-rhodes: As I re-read all these decref's, I'm confused about why they say things like `Py_DECREF(&SatrecType);` rather than `Py_DECREF(SatrecType);`. Isn't `SatrecType` already a pointer? With the `&` added, it looks like it would be trying to free the pointer rather than the object?

brandon-rhodes (changes requested):

brandon-rhodes:
Thanks so much for the forward progress on this! I've made three final comments which I think should resolve the remaining snags that I might see in the code.

extension/wrapper.cpp
neutrinoceros: Ah, that makes sense. Sorry I didn't think of that.

extension/wrapper.cpp
neutrinoceros: you're right. I don't know where this mistake came from. I probably was thinking compilation would just fail if I passed the wrong type. I'm actually not sure why it didn't. How is `PyObject**` acceptable where a `PyObject*` is expected ? (curious to know if you got an explanation for this, but I'll fix this in any case)

setup.py
neutrinoceros: Ah, a fellow Hettinbro !
*hat tip*

neutrinoceros:
Great, thank you so much for you careful review. I think all problems you spotted should be addressed now !

extension/wrapper.cpp
brandon-rhodes: I was asking myself exactly the same question as I was reading it! 🙂 I don't know why the compiler didn't flag it.

PR #155, opened 2026-03-11
BLD: avoid deprecation warnings from setuptools when building on Python 3

neutrinoceros:
This addopts the standard license metadata format defined in https://peps.python.org/pep-0639/, which recent versions of setuptools have been pressing users to migrate to, since it clashes directly with their equivalent schema.
I tried to preserve Python 2 compat while keeping duplications to a minimum.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,25 +59,30 @@ description, long_description = namespace['__doc__'].split('\n', 1)
 description = description.strip()
 long_description = long_description.strip() + '\n'
 
+if sys.version_info[0] == 2:
+    # keep in sync with the [project] table from pyproject.toml
+    setup_kwargs = {
+        "name": "sgp4",
+        "long_description": long_description,
+        "long_description_content_type": 'text/markdown',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -59,25 +59,30 @@
...
     url = 'https://github.com/brandon-rhodes/python-sgp4',
     classifiers = [
         'Development Status :: 5 - Production/Stable',
         'Intended Audience :: Science/Research',
-        'License :: OSI Approved :: MIT License',
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.8',
-        'Programming Language :: Python :: 3.9',
```
neutrinoceros, line 80: I'm removing these purely out of consistency with the existing `CIBW_SELECT` setting, but I note that I could probably easily re-introduce them later, virtually for free if #151 is accepted.

PR #156, opened 2026-03-11
RFC: restrict usage of `PyUnicode_*` functions to the Limited API subset

neutrinoceros:
needed for #151

This should be exactly equivalent. See [docs](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8)


neutrinoceros:
based off #153 for good measure.

PR #157, opened 2026-03-11
RFC: treat `PyTypeObject` as an opaque struct for Limited API compliance

neutrinoceros:
needed for #151
based off #153

This one doesn't compile yet, but the good news is that the errors are identical wether I force abi3 or not.

```
extension/wrapper.cpp:124:15: error: cannot initialize a variable of type 'allocfunc' (aka '_object *(*)(_typeobject *, long)') with an rvalue of type 'void *'
  124 |     allocfunc alloc_func = PyType_GetSlot(cls, Py_tp_alloc);
      |               ^            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
extension/wrapper.cpp:514:15: error: cannot initialize a variable of type 'allocfunc' (aka '_object *(*)(_typeobject *, long)') with an rvalue of type 'void *'
  514 |     allocfunc alloc_func = PyType_GetSlot(type, Py_tp_alloc);
      |               ^            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
```

### DIFF
--- a/extension/wrapper.cpp
+++ b/extension/wrapper.cpp
@@ -124,7 +124,8 @@ Satrec_twoline2rv(PyTypeObject *cls, PyObject *args)
     line2[68] = '\0';
 
     /* Allocate the new object. */
-    SatrecObject *self = (SatrecObject*) cls->tp_alloc(cls, 0);
+    allocfunc alloc_func = (allocfunc)PyType_GetSlot(cls, Py_tp_alloc);
+    SatrecObject *self = (SatrecObject*)alloc_func(cls, 0);
     if (!self)
         return NULL;
 
@@ -504,7 +505,9 @@ SatrecArray_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
     if (length == -1)
         return NULL;
 
-    return type->tp_alloc(type, length);
+    allocfunc alloc_func = (allocfunc)PyType_GetSlot(type, Py_tp_alloc);
+    PyObject *new_satrec = alloc_func(type, length);
+    return new_satrec;
 }
 
 


neutrinoceros:
The only difference I can see between this broken code and the known-working examples I have in astropy is that this module is C++ while the ones I can compile are C... if that's the meaningful difference I may be out of my depth, as I cannot think of a single reason why that would matter.

neutrinoceros:
found inspiration from [this stack overflow thread](https://stackoverflow.com/questions/24227723/c-cannot-initialize-variable-with-an-rvalue-of-type-void). It compiles okay if I use type casting. Maybe C compilers just do the exact same thing *implicitly* and I never knew it was needed ?

extension/wrapper.cpp
```diff
@@ -510,7 +511,9 @@ SatrecArray_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
     if (length == -1)
         return NULL;
 
-    return type->tp_alloc(type, length);
+    allocfunc alloc_func = (allocfunc)PyType_GetSlot(type, Py_tp_alloc);
+    PyObject *ret = alloc_func(type, length);
```
brandon-rhodes: Let's call this `new_satrec` instead of just `ret`. Thanks!

brandon-rhodes (changes requested):
I think I only spotted one naming tweak, then we can get this merged!

extension/wrapper.cpp
neutrinoceros: done

PR #159, opened 2026-03-12
MNT: move cibw's configuration to TOML

neutrinoceros:
Based off #155 which introduced `pyproject.toml`
Incidentally also includes #153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,12 +126,6 @@ jobs:
         uses: pypa/cibuildwheel@v3.3.1
         env:
           PYTHON_SGP4_COMPILE: always
-          # if adding python 3.15, just add cp315-*
-          CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-* cp314-*"
-          CIBW_BUILD_VERBOSITY: 0
-          CIBW_SKIP: '*-musllinux_*'
-          CIBW_TEST_REQUIRES: numpy
-          CIBW_TEST_COMMAND: python -m sgp4.tests
 
       - name: upload wheels
         uses: actions/upload-artifact@v6
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,5 +11,21 @@ name = "sgp4"
 readme = "README.md"
 license = "MIT"
 license-files = ["LICENSE"]
-requires-python = ">=2.7,!=3.0,!=3.1,!=3.2,!=3.3,!=3.4,!=3.5,!=3.6,!=3.7,!=3.8,!=3.9"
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -125,12 +125,7 @@ jobs:
       - name: Build wheels python 3.10 - 3.14
         uses: pypa/cibuildwheel@v3.3.1
         env:
-          # if adding python 3.15, just add cp315-*
-          CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-* cp314-*"
```
neutrinoceros, line 130: I intentionally didn't reproduce this line verbatim because cibw's already parses `project.requires-python` from `pyproject.toml`, and set the appropriate upper bound itself.

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[dependency-groups]
+dev = [
+    "numpy>=1.21.2",
+]
+
+[tool.cibuildwheel]
+build-verbosity = 0
+skip = [
+    "*-musllinux_*", # not sure there's a good reason for this anymore ?
```
neutrinoceros, line 25: very likely this was done because numpy didn't serve this target at the time, but it does now, so maybe we can re-enable it here or in a follow up PR ?

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+    "cp314t-*",
```
neutrinoceros, line 26: recent versions of cibw include Free-threading 3.14 (3.14t) by default (though, not 3.13t). I assume we don't want to add it here.
FWIW it might get significantly easier to serve both builds (3.1x and 3.1xt) with a single wheel per platform in the near future, provided PEP 803 is accepted for Python 3.15

pyproject.toml
brandon-rhodes: Yes, let's save that for a future PR. Thanks for pointing it out; in the original commit that added it, I don't see a comment from the contributor explaining why, so we're left with the guess that it was NumPy.

pyproject.toml
brandon-rhodes: I agree that this would not be the PR for introducing a free-threading build, and that it might be useful to hold off and see what happens with PEP 803. I'll have to go read it sometime. Thanks!

brandon-rhodes:
I like how this pulls project data out of a proprietary CI YAML file, and into the main project configuration. Thanks for working on it!

pyproject.toml
neutrinoceros: PEP 803 is now accepted and implementation is expected to land in 3.15.0b1, FYI

pyproject.toml
neutrinoceros: #161 👀

PR #170, opened 2026-09-01
Support OMM NORAD catalog numbers beyond Alpha-5 range (closes #169)

### LINKED ISSUE #169: Support OMM NORAD catalog numbers beyond the Alpha-5 range
Hello,

as discussed in skyfielders/python-skyfield#995, OMM data can contain NORAD catalog numbers larger than the range supported by TLE Alpha-5 encoding.

I ran into the same limitation when initializing a `Satrec` from OMM data:

```python
fields["NORAD_CAT_ID"] = 999100101

satrec = Satrec()
omm.initialize(satrec, fields)
```

This raises:

```text
ValueError: satellite number cannot exceed 339999, whose Alpha 5 encoding is 'Z9999'
```

I think it would be useful if `omm.initialize()` could support the full catalog number allowed by OMM, independently of the TLE/Alpha-5 limitation. As far as I understand, the catalog number is identification metadata and does not affect SGP4 propagation.

If larger catalog numbers are supported by `Satrec`, this also raises the question of `export_tle()`. A `Satrec` could then contain a catalog number that cannot be represented in a TLE.

Maybe `export_tle()` could continue to raise by default in that case, but optionally accept a fallback satellite number, for example:

```python
export_tle(satrec, fallback_satnum=0)
```

The fallback would only be used when the real catalog number cannot be represented by Alpha-5. Without the argument, the current error behavior could remain unchanged.

That would allow callers that need a TLE for software with legacy format requirements to explicitly choose a placeholder, while avoiding silently replacing the real catalog number.

Greetings,
Michael

karlhillx:
Closes #169

## Problem

`omm.initialize()` raised `ValueError: satellite number cannot exceed 339999` for any NORAD_CAT_ID greater than 339999. This is a problem because OMM (CCSDS 502.0) is a modern XML/CSV/JSON format that carries the catalog ID as a plain integer — not the 5-character TLE column where the Alpha-5 encoding is needed to fit 6-digit numbers into 5 columns.

Real-world examples that hit this:
- Newer satellites with IDs in the 800000–999999 range
- Recently reclassified debris objects

The bug reproduced with `NORAD_CAT_ID=999999` against `master`.

## Root cause

`omm.initialize()` passed `NORAD_CAT_ID` (an int) directly to `sgp4init()`, which calls `to_alpha5()` whenever its `satn` argument is an int. `to_alpha5()` can only encode satnums up to 339999 (whose Alpha-5 form is `'Z9999'`) and raises for anything larger.

The TLE path never had this problem — `twoline2rv()` extracts `satrec.satnum_str` directly from the 5-character column in the TLE and passes it as a string to `sgp4init()`, bypassing `to_alpha5()` entirely.

## Fix

In `sgp4/omm.py`, convert the integer NORAD_CAT_ID to a string before passing it to `sgp4init()`. The string takes the `if isinstance(satn, int):` branch into a no-op, so `to_alpha5()` is skipped. The full integer round-trips correctly because `from_alpha5()` decodes any digit-led string as `int(s)`.

The TLE code path is unchanged.

## Behavior after the fix

Verified for satnums in the range [5, 9999999]:

| satnum | satnum_str  | sgp4(epoch) |
|-------:|-------------|-------------|
|      5 | `'5'`       | OK          |
|  55123 | `'55123'`   | OK (MARIO)  |
| 339999 | `'339999'`  | OK (Alpha-5 max) |
| 340000 | `'340000'`  | OK (previously raised) |
| 999999 | `'999999'`  | OK (the issue's example) |
|9999999 | `'9999999'` | OK (seven digits) |

Note: in the OMM path, `satnum_str` is the raw digit string (no zero-padding), which is fine for round-trip and propagation but means `export_tle()` cannot represent such a satrec back as a TLE. That's an inherent limitation of the TLE format itself (the column is exactly 5 characters), not a regression.

## Tests

Added two tests in `sgp4/tests.py`:

- `test_omm_supports_satnum_beyond_alpha5_range` — exercises the reported case (`NORAD_CAT_ID=999999`) and asserts both `sat.satnum` and `sat.satnum_str` round-trip correctly.
- `test_omm_satnum_at_alpha5_boundary` — checks both 339999 (Alpha-5 max, the legacy encoding boundary) and 340000 (one past, the previously broken case) via the OMM path.

`test_satnum_that_is_too_large` still raises `ValueError` for direct `sat.sgp4init(int)` calls with int > 339999, preserving the existing API contract for callers that go through the legacy int→Alpha-5 path.

## Test results

```
$ PYTHON_SGP4_COMPILE=never python3 -m unittest sgp4.tests
...............................................
Ran 47 tests in 0.053s

OK
```

47 tests pass (45 pre-existing + 2 new). Pre-existing pyflakes warning on `sgp4/conveniences.py:9` is unchanged.

## Diff stat

```
sgp4/omm.py   |  8 +++++++-
sgp4/tests.py | 28 ++++++++++++++++++++++++++++
2 files changed, 35 insertions(+), 1 deletion(-)
```

### DIFF
--- a/NOTES.txt
+++ b/NOTES.txt
@@ -1,5 +1,0 @@
-
-After doing a release, I should also publish a source wheel:
-
-PYTHON_SGP4_COMPILE=never PYTHONDONTWRITEBYTECODE= uv build
-uv publish dist/...
--- a/sgp4/omm.py
+++ b/sgp4/omm.py
@@ -47,5 +47,11 @@ def initialize(sat, fields, gravconst=WGS72):
     nodeo = float(fields['RA_OF_ASC_NODE']) * _to_radians
     satnum = int(fields['NORAD_CAT_ID'])
 
-    sat.sgp4init(gravconst, 'i', satnum, epoch, bstar, ndot, nddot, ecco,
+    # OMM (CCSDS 502.0) is a modern XML/CSV/JSON format that carries the
+    # NORAD catalog ID as a plain integer, not the 5-character TLE column.
+    # Pass the number as a string so that sgp4init() skips `to_alpha5()`,
+    # which can only encode satnums up to 339999 ('Z9999'). The full
+    # integer round-trips correctly through `from_alpha5()` because a
+    # digit-led string is decoded as a plain `int(s)`.
+    sat.sgp4init(gravconst, 'i', str(satnum), epoch, bstar, ndot, nddot, ecco,
                  argpo, inclo, mo, no_kozai, nodeo)
--- a/sgp4/omm.py
+++ b/sgp4/omm.py
@@ -47,11 +47,5 @@ def initialize(sat, fields, gravconst=WGS72):
     nodeo = float(fields['RA_OF_ASC_NODE']) * _to_radians
     satnum = int(fields['NORAD_CAT_ID'])
 
-    # OMM (CCSDS 502.0) is a modern XML/CSV/JSON format that carries the
-    # NORAD catalog ID as a plain integer, not the 5-character TLE column.
-    # Pass the number as a string so that sgp4init() skips `to_alpha5()`,
-    # which can only encode satnums up to 339999 ('Z9999'). The full
-    # integer round-trips correctly through `from_alpha5()` because a
-    # digit-led string is decoded as a plain `int(s)`.
-    sat.sgp4init(gravconst, 'i', str(satnum), epoch, bstar, ndot, nddot, ecco,
+    sat.sgp4init(gravconst, 'i', satnum, epoch, bstar, ndot, nddot, ecco,
                  argpo, inclo, mo, no_kozai, nodeo)
--- a/sgp4/tests.py
+++ b/sgp4/tests.py
@@ -857,6 +857,34 @@ def test_omm_uses_gravconst_parameter():
     assertEqual(sat1.mu, 398600.8)
     assertEqual(sat2.mu, 398600.5)
 
+def test_omm_supports_satnum_beyond_alpha5_range():
+    # https://github.com/brandon-rhodes/python-sgp4/issues/169
+    # The Alpha-5 encoding used by TLEs caps NORAD_CAT_ID at 339999.
+    # OMM is a modern format (CCSDS 502.0) and can carry larger integers,
+    # so the OMM init path must accept them without raising.
+    MARIO_LARGE_CSV = """\
+OBJECT_NAME,OBJECT_ID,EPOCH,MEAN_MOTION,ECCENTRICITY,INCLINATION,RA_OF_ASC_NODE,ARG_OF_PERICENTER,MEAN_ANOMALY,EPHEMERIS_TYPE,CLASSIFICATION_TYPE,NORAD_CAT_ID,ELEMENT_SET_NO,REV_AT_EPOCH,BSTAR,MEAN_MOTION_DOT,MEAN_MOTION_DDOT
+MARIO,1998-067UQ,2023-04-25T10:45:30.642912,15.99081912,.0014649,51.6242,216.2930,331.8976,28.1241,0,U,999999,999,1839,.1568E-2,.787702E-2,.29408E-3
+"""
+    fields = next(omm.parse_csv(StringIO(MARIO_LARGE_CSV)))
+    sat = Satrec()
+    omm.initialize(sat, fields)
+    assertEqual(sat.satnum, 999999)
+    # The raw integer is stored verbatim in satnum_str; from_alpha5()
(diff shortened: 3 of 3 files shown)


hneogy:
Thanks for picking this up. I tested the PR at commit 5e4f308 against a conformance corpus for the
2026 catalog-number migration (https://github.com/hneogy/gp-omm-conformance), on Python 3.14.4 with
both the accelerated and the pure-Python `Satrec`.

The two new tests are useful and target the defect exactly. One thing worth flagging: at this
commit, the `omm.initialize` change from the first commit is no longer present (the second commit's
message says it was removed for C++ accelerated compatibility), so the tests are currently the only
change.

Because both tests skip when `api.accelerated` is true, a green CI run on the accelerated build
would not show whether the nine-digit case loads.

What I saw at 5e4f308:

- The CCSDS 502.0-B-3 annex G example OMM with `NORAD_CAT_ID` 799501621 still raises
  `ValueError: satellite number cannot exceed 339999` on both builds, as in 2.27.
- On the pure-Python build, both new tests fail with that error.
- The corpus's 16 cases give the same results as 2.27: no fixes, no regressions.

Happy to test any follow-up commit against live nine-digit data and the corpus.

karlhillx:
Thanks for this — testing against the conformance corpus on both builds is exactly the review this needed, and you caught something the CI skip condition would have hidden.

Your reading is correct. Commit `5dd00409` removed the `omm.initialize()` change from `95f7f4e9` because the accelerated `Satrec.sgp4init()` (in `extension/wrapper.cpp`) only accepts an integer satnum and carries its own Alpha-5 range check — the str-passing approach failed outright under the C++ build, so the fix came out rather than ship something that only worked on one backend.

Which leaves the PR at HEAD as tests-only, as you found: it documents the defect but doesn't fix #169, the new tests fail on pure-Python, and the skip message is now stale (it describes the removed str-passing approach). The Copilot review has the same staleness, since it was filed against commit 1.

The follow-up fix has to live at the backend, not in `omm.py`, and it needs to be conditional. Two things the next commit has to get right that commit 1 didn't:

1. **Scope the bypass to out-of-range IDs only.** Copilot flagged this against the first-commit diff — unconditionally passing `str(satnum)` would alter `satnum_str` for every OMM init, breaking `export_tle()` round-trip and existing `satnum_str` expectations for in-range IDs. The fix should pass `int(satnum)` normally and only switch to the bypass path when `satnum > 339999` (or whatever threshold the new backend handling settles on).

2. **Touch both backends.** The 339999 cap is enforced in two places: `sgp4/alpha5.py` (pure-Python, bypassed by a digit-led string) and `extension/wrapper.cpp` (the C++ accelerated path, which doesn't accept strings at all). A complete fix relaxes the wrapper's handling so large integer satnums are stored verbatim in `satnum_str`, after which `omm.py` can stay on the integer path for everything — no string detour needed.

That's a change to the compiled extension, so I'd like the maintainer's read on the approach before pushing it — it also interacts with the `export_tle()` question from #169 (a `Satrec` holding a satnum that can't round-trip through Alpha-5). I'll rework the PR once the design is settled and would genuinely appreciate a re-run against the corpus once it's up — your nine-digit CCSDS annex G case is the right acceptance test.

brandon-rhodes:
So. The `struct elsetrec` provides only a tiny bit of storage for satellite numbers:
```
  char      satnum[6];
```
And so the code in this library engages in the contortions specified by alpha5 to get larger satellite numbers stored in that field. And can't store larger numbers at all.

1. What if I just abandoned the idea of compiling Vallado's original code, and bumped the field out to `satnum[20]` or something extravagant like that?
2. Or, up at the Python level, we could look at `SatrecObject`, which currently has only one field, the `elsetrec satrec;`. We could add a second field for the satellite number that supports storing huge integers. Even just `unsigned int satnum` or something, which on most platforms supports quite a few digits.
3. Or, we could keep restricting storage to the little `satnum[6]`, but use a pure binary storage for integers. Six bytes should successfully store satellite IDs through 281474976710655.

Any of these three tweaks would let huge satellite integers be accepted from OMM and then fed back out to OMM, and also be set through `.satnum` and accessed through `.satnum`, without further problems. Only the TLE import path and TLE export path would need to worry about alpha5 or about running into IDs that are too large to process. (Oh — maybe work would also need to be done so that the attribute `.satnum_str` also looked alpha-5-ish? I'm not clear on how users expect that field to behave, or what they use it for.)

As I've not touched this library since its most recent release back in July, it's possible that I've already shared these ideas on other open issues? I suppose that, tomorrow morning, I should go through all the open issues and figure out how many places satellite number discussions are currently underway.

PR #172, opened 2026-09-21
Accept an empty or `UNKNOWN` OBJECT_ID in `sgp4.omm`

### LINKED ISSUE #171: `sgp4.omm.parse_xml` + `initialize` raise `TypeError` on an empty `<OBJECT_ID/>` (CelesTrak analyst objects)
## Environment

- python-sgp4 2.27 (the latest release on PyPI as of 2026-09-21), Python 3.14.4, macOS

## Summary

Related but distinct from #168, which is about `export_omm()` on an already-loaded `Satrec` with an empty designator; this report is about *loading* such a record from XML, where the empty element arrives as `None` rather than `''`.

CelesTrak's OMM XML for analyst objects (objects with no international designator) writes an
empty element `<OBJECT_ID></OBJECT_ID>`. `sgp4.omm.parse_xml` copies `element.text`, which
`xml.etree.ElementTree` returns as `None` for an empty element, and `sgp4.omm.initialize`
then evaluates `fields['OBJECT_ID'][2:]`:

```
TypeError: 'NoneType' object is not subscriptable
```

Scale, from live data on 2026-09-21: `https://celestrak.org/NORAD/elements/gp.php?GROUP=analyst&FORMAT=XML`
held 565 records, 563 of them with an empty `OBJECT_ID`; **563 of 565 records cannot be
loaded through the XML path**, while the identical records fetched as CSV
(`...&FORMAT=CSV`, where the empty field is the empty string `''`) load without error.
The first-ever record for analyst id 270449 (`gp-first.php?CATNR=270449&FORMAT=XML`) fails
the same way.

CelesTrak documents the behaviour: "an object in the current analyst sat range (80000-series)
typically will not have a name (OBJECT_NAME) or International Designator (OBJECT_ID)" and
"Some elements may be blank (KVN) or null (XML)"
(https://celestrak.org/NORAD/documentation/gp-data-formats.php). Space-Track defines analyst
objects as tracked "with insufficient fidelity for publication in the public satellite
catalog" (https://www.space-track.org/documentation). Note that CCSDS 502.0-B-3 Table 4-2
recommends the literal `UNKNOWN` for an unknown `OBJECT_ID`; CelesTrak emits an empty element
instead, so parsers meet both.

## Minimal standalone reproducer (no network)

Element values are the CCSDS 502.0-B-3 annex G example; the document shape (an `<ndm>`
wrapper, OMM version 2.0, empty header elements, empty `OBJECT_ID`) is CelesTrak's.

```python
import io
from sgp4.api import Satrec
from sgp4 import omm

xml = """<?xml version="1.0" encoding="UTF-8"?>
<ndm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://sanaregistry.org/r/ndmxml_unqualified/ndmxml-2.0.0-master-2.0.xsd">
<omm id="CCSDS_OMM_VERS" version="2.0">
<header><CREATION_DATE/><ORIGINATOR/></header><body><segment><metadata><OBJECT_NAME>UNKNOWN</OBJECT_NAME><OBJECT_ID></OBJECT_ID><CENTER_NAME>EARTH</CENTER_NAME><REF_FRAME>TEME</REF_FRAME><TIME_SYSTEM>UTC</TIME_SYSTEM><MEAN_ELEMENT_THEORY>SGP4</MEAN_ELEMENT_THEORY></metadata><data><meanElements><EPOCH>2020-03-04T10:34:41.426400</EPOCH><MEAN_MOTION>1.00273272</MEAN_MOTION><ECCENTRICITY>.0005013</ECCENTRICITY><INCLINATION>3.0539</INCLINATION><RA_OF_ASC_NODE>81.7939</RA_OF_ASC_NODE><ARG_OF_PERICENTER>249.2363</ARG_OF_PERICENTER><MEAN_ANOMALY>150.1602</MEAN_ANOMALY></meanElements><tleParameters><EPHEMERIS_TYPE>0</EPHEMERIS_TYPE><CLASSIFICATION_TYPE>U</CLASSIFICATION_TYPE><NORAD_CAT_ID>81011</NORAD_CAT_ID><ELEMENT_SET_NO>999</ELEMENT_SET_NO><REV_AT_EPOCH>4316</REV_AT_EPOCH><BSTAR>.1E-3</BSTAR><MEAN_MOTION_DOT>-.113E-5</MEAN_MOTION_DOT><MEAN_MOTION_DDOT>0</MEAN_MOTION_DDOT></tleParameters></data></segment></body></omm>
</ndm>
"""

fields = next(omm.parse_xml(io.StringIO(xml)))
print(repr(fields["OBJECT_ID"]))     # None
omm.initialize(Satrec(), fields)     # TypeError: 'NoneType' object is not subscriptable
```

Expected: the record loads with an empty international designator (`intldesg == ''`), as it
does from CSV. Actual: `TypeError`.

## Live reproducer (one request; respect CelesTrak's usage policy)

```python
import io, urllib.request
from sgp4.api import Satrec
from sgp4 import omm

url = "https://celestrak.org/NORAD/elements/gp.php?GROUP=analyst&FORMAT=XML"
req = urllib.request.Request(url, headers={"User-Agent": "sgp4-issue-repro (single request)"})
text = urllib.request.urlopen(req, timeout=60).read().decode()
ok = bad = 0
for fields in omm.parse_xml(io.StringIO(text)):
    try:
        omm.initialize(Satrec(), fields); ok += 1
    except TypeError:
        bad += 1
print(ok, "loaded;", bad, "TypeError")    # 2 loaded; 563 TypeError on 2026-09-21
```

## Suggested fix, and an offer

In `parse_xml`, use `(field.text or '')`; in `initialize`, derive `intldesg` only from an
`OBJECT_ID` of the form `YYYY-NNNP...` and otherwise set it to `''` (this also covers the
CCSDS-recommended literal `UNKNOWN`, which today yields `intldesg == 'KNOWN'`). The same guard
applies to `OBJECT_NAME`, which the provider says can also be blank.

I have a patch ready with two tests (`test_omm_xml_with_empty_object_id`,
`test_omm_unknown_object_id`) in the style of `sgp4/tests.py`; it applies cleanly to master
and the existing test module still passes with it. Happy to open a pull request if that is
welcome, or to adjust the approach if you would rather handle it differently.

Found while building a public conformance corpus for the catalog-number migration: https://github.com/hneogy/gp-omm-conformance

hneogy:
Fixes #171

CelesTrak's OMM XML for analyst objects has an empty `<OBJECT_ID></OBJECT_ID>`. ElementTree gives `None`
for the empty element, `parse_xml()` passes it through, and `initialize()` raises `TypeError` on
`fields['OBJECT_ID'][2:]`. The same records in CSV, where the value is `''`, load fine.

What changed:

- `parse_xml()` stores an empty element as `''`, so an empty `OBJECT_ID` (or `OBJECT_NAME`) arrives the
  same way from XML as from CSV.
- `initialize()` derives `intldesg` only from an `OBJECT_ID` of the form `YYYY-NNNP...`; otherwise it
  sets `''`. That also covers the CCSDS-recommended literal `UNKNOWN`, which used to produce
  `intldesg == 'KNOWN'`.

Tests: `test_omm_xml_with_empty_object_id` and `test_omm_unknown_object_id` in `sgp4/tests.py`, next to
the other OMM tests. Both fail on master and pass with the change.

Verified with `python -m sgp4.tests` on Python 3.14 (macOS): 47 tests pass on both the pure-Python build
(`PYTHON_SGP4_COMPILE=never`) and the compiled C++ build; the README doctests pass too.


sgp4/omm.py
```diff
@@ -28,7 +28,13 @@ def parse_xml(file):
 
 def initialize(sat, fields, gravconst=WGS72):
     sat.classification = fields['CLASSIFICATION_TYPE']
-    sat.intldesg = fields['OBJECT_ID'][2:].replace('-', '')
+    object_id = fields.get('OBJECT_ID') or ''
```
brandon-rhodes, line 32: Fun trivia: this can be slightly tighter and faster, since `.get()` takes, as its second argument, the default value to use if `None` isn't the preferred default 🙂

Thus, `fields.get('OBJECT_ID', '')`

sgp4/omm.py
```diff
@@ -28,7 +28,13 @@ def parse_xml(file):
...
+    # Analyst objects have no international designator: CelesTrak writes an
+    # empty OBJECT_ID; CCSDS 502.0-B-3 permits the literal UNKNOWN.
+    if len(object_id) > 4 and object_id[4] == '-':
```
brandon-rhodes: If you wanted to skip the `and`, you could say `if object_id[4:5] == '-':`, since the slice will succeed and return the empty string if `object_id` isn't long enough. But I'll let you decide whether that relies on a too-obscure feature of Python or not. 🙂

brandon-rhodes (commented):
Thanks, this looks good! Here are a couple of tweaks to consider, then we can get this landed.

sgp4/omm.py
hneogy: Good tip, thanks — I didn't know .get() took the default as a second argument.

The reason I'd reached for `or ''` is that None can still land here even with parse_xml() now storing empty elements as '': a JSON record with "OBJECT_ID": null passed straight in (the README's json.load route), or a short CSV row via DictReader's restval. In those cases .get('OBJECT_ID', '') returns None and object_id[4:5]
raises the same TypeError this PR fixes. So I've kept `or ''` and added a one-line comment.

That said, it's your call whether callers passing null deserve the error — happy to switch to your form if so.

sgp4/omm.py
hneogy: Took it, thanks. I don't think it's too obscure — a slice that returns '' on a short string is standard Python, and it reads cleaner than the length check. In as of the push.

hneogy:
Pushed both tweaks — took the slice as suggested, and kept `or ''` for the None case for the reason in the thread above. Thanks for the review; ready for another look whenever you have a moment.

sgp4/omm.py
brandon-rhodes: For now, let's tighten it up, since it makes the type-contract clearer, in case I ever add type hints to this package: input keys and values are all expected to be strings. I have not yet encountered, in the wild, either of the two interesting cases you outline (JSON null, or short CSV lines producing None values), and if they wind up being a problem for folks, then we can reconsider later.

(I like, by the way, how down in the tests you followed the pattern of simply taking existing examples, and replacing the one piece that needed tweaking to produce the issue under test; rather than having whole new satellites, just to try out one behavior. Nice!)

hneogy: Makes sense — switched to .get('OBJECT_ID', '') and dropped the comment, since it no longer applies. The type contract is a good reason, and you're right that neither case has actually turned up in the wild. If one ever does I'll open a separate issue rather than reopen this.

And thank you — that's kind of you to say.

brandon-rhodes:
Thanks for putting these tweaks together!

